PR #4, opened 2016-09-23
Adding Container example

Banashek:
Here's a simple container example for react-native showing how two independent counters can be composed into a container view.


Banashek:
Only tested these on ios, but they should work on Android.

PR #5, opened 2016-09-24
Adding async web-request example

Banashek:
Another example (just adding examples as I figure out how to do various things haha). It's almost like documentation for myself too.


et1975:
Thanks for doing this, I'm away from computer but I'll merge it today.

Banashek:
No problem. Sounds good.

et1975:
@Banashek here's slightly nicer syntax to go about it: https://github.com/et1975/fable-elmish/blob/dev/samples/react-native/parent-child/src/Main/State.fs#L28
You just need to include `fable-promise.fs` and then the `Cmd.ofPromise` becomes available.
I'm reorganizing the namespaces a bit, so up to you if you want try it as is or wait until new elmish is out.

Banashek:
I was actually just putting together a simple subscriber example that looks almost exactly like that:
```
module AwesomeApp

open System
open Fable.Core
open Fable.Import
open Fable.Import.ReactNative
open Fable.Helpers.ReactNative
open Fable.Helpers.ReactNative.Props

open Fable.Elmish

// Model

type Msg =
  | Increment
  | Decrement
  | NoOp

type Model = 
  { count:int
    currentMsg:Msg }

let init () =
  { count = 0
    currentMsg = NoOp }, []

// Update

let update (msg:Msg) (model:Model) : Model*Cmd<Msg> =
  match msg with
  | Increment ->
    { model with
        count = model.count + 1
        currentMsg = Increment }, []
  | Decrement ->
    { model with
        count = model.count - 1
        currentMsg = Decrement }, []
  | NoOp ->
      model, []

// Subscriptions
let timerTick model dispatch =
  let t = new Timers.Timer 1000.
  t.Elapsed.Subscribe(fun _ -> dispatch (model.currentMsg: Msg)) |> ignore
  (*t.Elapsed.Subscribe(fun _ -> dispatch Increment) |> ignore*)
  t.Enabled <- true

let subscription model =
  Cmd.ofSub (timerTick model)


// rendering views with ReactNative
module R = Fable.Helpers.ReactNative

let view {count = count} dispatch =
  let onClick msg =
    fun _ -> msg |> dispatch 

  R.view []
    [ Styles.button "-" (onClick Decrement)
      R.text [] (string count)
      Styles.button "+" (onClick Increment)
    ]

// App
let program =
    Program.mkProgram init update
    |> Program.withSubscription subscription
    |> Program.withConsoleTrace

type App() as this =
    inherit React.Component<obj, Model>()

    let safeState state =
        match unbox this.props with 
        | false -> this.state <- state
        | _ -> this.setState state

    let dispatch = program |> Program.run safeState

    member this.componentDidMount() =
        this.props <- true

    member this.render() =
        view this.state dispatch
```

You can see I was trying to figure out a way to dynamically update a subscription during the apps lifecycle.

The intended functionality was for the last message sent (Increment or Decrement) would be sent again every second.

Works fine if I initialize it with Increment or Decrement, but not when using the model property. Makes sense though since it's initialized once.

Why the namespace reorg? I'm sure it's for a good reason, just curious =).

Banashek:
Swapped the async to promise

samples/react-native/webrequest/src/App.fs
```diff
@@ -5,6 +5,9 @@ open Fable.Elmish
 open Fable.Import.Fetch
 open Fable.Helpers.Fetch
 open Fable.Core.Extensions
+open Fable.Extras.Promise
```
et1975: Just opening `Fable.Extras` is good enough. I should actually mark `Cmd` module us unopable, because I want no ambiguity about it...

samples/react-native/webrequest/src/App.fs
```diff
@@ -0,0 +1,103 @@
...
+
+// Update
+let update msg model : Model*Cmd<Msg> =
+  match msg with
+  | ChangeInput str ->
+    { model with Url = str}, []
+  | SendRequest ->
+    { model with
+        StatusStr = (statusString InProgress)
+        Status = InProgress }, Fable.Extras.Cmd.ofPromise (fun _ -> getUrl model.Url) Success Error
```
et1975: Looks like I need to make `ofPromise` cleaner...

et1975: As I started going beyond basics I realized names like `Fable.Extras` just don't make sense. Not sure `Fable` prefix adds a value either, one could use the core abstractions with any runtime.

Banashek: I agree. Using the "Elmish" namespace might make more sense. 

Elmish.Core
Elmish.Extras

seems a bit more appropriate.

PR #8, opened 2016-09-30
make build.cmd work

forki:


forki:
Cmd.ofPromise getUrl model.Url Success Error

mhm this would not work after this change. Maybe it needs to be rewitten!?

et1975:
As discussed, works with a newer F# compiler, closing for now. Not sure I want `npm` as nuget, but I might be convinced with a good argument.

PR #13, opened 2016-10-28
Change ofAnimationFrame to more generic ofCallback

davidtme:
This still allows you to use requestAnimationFrame:
```
Cmd.ofCallback (fun _ c -> Browser.window.requestAnimationFrame(fun _ -> c()) |> ignore) () (fun _ -> Advance) (fun _ -> invalidOp "Can't happen")
```
But doesn't limit it to just requestAnimationFrame.

### DIFF
--- a/src/elmish/elmish.fs
+++ b/src/elmish/elmish.fs
@@ -41,18 +41,17 @@ module Cmd =
             }
         [bind >> Async.StartImmediate]
 
-    /// Command that will schedule animation frame callback and map the result 
-    /// into success or error (of exception) 
-    let ofAnimationFrame (task:float->_) (ofSuccess:_->'msg) (ofError:_->'msg) : Cmd<'msg> =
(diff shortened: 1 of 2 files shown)


et1975:
Thanks for the PR!

However:
- I think considering the timer nature of `requestAnimationFrame` and the fact that it's browser-specific, I'm going to drop it from the API. 
- `ofCallback` usage looks too cumbersome to be in the base Cmd API, I think you'll be better off by implementing it in your project.

davidtme:
@et1975 your Cmd.ofAnimationFrame pointed my in the right direction so thank yous for your help.

Just in case anyone else needs it this is how I ended up using requestAnimationFrame

```F#
let update message model =
    match message with
    | Advance ->
        model, [fun dispatch -> Browser.window.requestAnimationFrame (fun _ ->  Advance |> dispatch) |> ignore]

```

PR #14, opened 2016-10-28
Add runWithView

davidtme:
This allow more ways to use elmish including being able to use the `view` part of the `program`


et1975:
Thanks for the PR, but I'm going to decline this change. I think I understand the motivation, but it takes it outside of the intended use.

That shouldn't stop you from running your own implementation of dispatch loop on top of the elmish abstractions though.

et1975:
I think I want this after all, thanks for the PR!
Note, I might tweak things a bit, it will be a breaking release.

davidtme:
@et1975 ace, I can get rid of my homebrew dispatch loop

PR #20, opened 2016-11-22
Update Fable version

forki:


et1975:
@forki hey, sorry but I'm going to close this w/o merging - the package.json changes you are overwriting are intentional, although I should probably declare fable-core as a `peerDependency`. Also, when 0.7 branch is ready I'll merge it into master, so there's no need to bring in master changes into it.

PR #23, opened 2017-01-11
very lame attempt to fix #22

### LINKED ISSUE #22: Does not build on Windows
When I cloned the repo, it does not build on my machine where otherwise [I use fable](http://www.naseukoly.cz/blog/2016/12-20-hello-fable-import-sharepoint/index.html).

I am pretty sure it is because I miss something and generally I am too pedantic on the DevOps (does that still exist?).

	Building: C:\Users\david.podhola\Documents\S\fable-elmish\src\elmish
	C:\Users\david.podhola\Documents\S\fable-elmish\packages\Npm.js\tools\npm.cmd run build
	Running build failed.
	Error:
	System.Exception: Start of process C:\Users\david.podhola\Documents\S\fable-elmish\packages\Npm.js\tools\npm.cmd failed. The system cannot find the file specified

capture

davidpodhola:
fixing only `build.cmd`, not `build.sh`. `npm install` should be probably called in `build.fsx` the same way I did it for the projects.

Not optimized at all, `npm install` should be probably not called when `node_modules` folder is present (?)

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -11,6 +11,4 @@ if errorlevel 1 (
   exit /b %errorlevel%
 )
 
-call packages\Npm.js\tools\npm.cmd install
-
 packages\FAKE\tools\FAKE.exe build.fsx %*
--- a/build.cmd
+++ b/build.cmd
@@ -11,4 +11,6 @@ if errorlevel 1 (
   exit /b %errorlevel%
 )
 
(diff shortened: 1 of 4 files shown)


davidpodhola:
Please do not merge yet, I forgot `samples`.

davidpodhola:
Already works

PR #32, opened 2017-02-12
Re-enable persistent TODO list

MaxWilson:
Re-enable persistence in TodoMVC example. Local storage must use toJson/ofJson instead of JSON.stringify/parse so that lists will still deserialize as iterable.


et1975:
Thanks

PR #33, opened 2017-02-14
Minor spacing updates

dsyme:
These normalize some spacing and notation to more standard F#

I didn't compile this, so please check.


et1975:
Thanks!

PR #36, opened 2017-02-28
[READY] Add snabbdom renderer

MangelMaxime:
**Not ready to merge**

## Description 
This PR add a new renderer: Elmish-Snabbdom [Snabbdom](https://github.com/snabbdom/snabbdom)

## Status
Currently the renderer is working, but we should improve the DSL and provide helpers where needed.

I will implement the same samples as for Elmish-react. Like so, we should have some basics to test the DSL and adapt it.

## How is it working ?
The DSL is a direct mapping over snabbdom hyperscript functions via `KeyValueList`


MangelMaxime:
One improvement that could be done, is support void element.

For example:

```fsharp
// Current version
[<Emit(typeof<Emitter>, "Tagged", "br")>]
let br b c = hyperscript "br" b c

// Proposition
[<Emit(typeof<Emitter>, "TaggedVoid", "br")>]
let br b = hyperscriptVoid "br" b
```

As the **br** element is a self-closing tag we can't at child in it. And it's noisy to do `br [] []`

alfonsogarciacaro:
@MangelMaxime Great work! :clap:

About the `br` tag, I also realized that some elements cannot have children in React, but I've kept the second argument in order not to break existing code. You've an opportunity here to fix that mistake ;)

Because the `Tagged` method of the emitter is already taking only one argument (plus the tag) in consideration, you shouldn't need to do anything special for that. This should work:

```fsharp
[<Emit(typeof<Emitter>, "Tagged", "br")>]
let br b: Element = jsNative
```

> Please replace `Element` with the proper return type of `br`

MangelMaxime:
Oh ok thanks for the explanations @alfonsogarciacaro

Zaid-Ajaj:
What is this PR missing to fully support Snabbdom in elmish?

MangelMaxime:
This is still a WIP.

MangelMaxime:
LIttle update.

The emitter has been fixed. And the DSL has been enhanced (new Events name + helpers for classes usages).
And navigation sample is working with Snabbdom renderer.
I will work on Todo sample tomorrow.

MangelMaxime:
@et1975 You can merge this PR when ready.
I already publish the `0.1.0` version of the package on npm.

samples/snabbdom/counter/README.md
```diff
@@ -0,0 +1,10 @@
+This is a port of [Elm's counter example](https://github.com/evancz/elm-architecture-tutorial/blob/master/examples/1-button.elm) implemented in F# and targeting Fable and React.
```
et1975: ~React~
snabbdom?

samples/snabbdom/counter/webpack.config.js
```diff
@@ -0,0 +1,59 @@
...
+    'webpack/hot/only-dev-server',
+    "./out"
+  ];
+  cfg.plugins = [
+    new webpack.HotModuleReplacementPlugin()
+  ];
+  cfg.module.loaders = [{
+    test: /\.js$/,
+    exclude: /node_modules/,
+    loader: "react-hot-loader"
```
et1975: react-specific plugin

MangelMaxime: You mean all the 4 lines or only the loader ?
Should we remove ?
```js
    'webpack/hot/only-dev-server',
```

and 

```js
  cfg.plugins = [
    new webpack.HotModuleReplacementPlugin()
  ];
```

Because we don't support HMR for Snabbdom I think.

et1975: I just meant react-hot-loader, not sure about the rest of the HMR, but it seems like it should work with anything, but maybe @alfonsogarciacaro could chip in here.

alfonsogarciacaro: The react-hot-loader is continuously changing but the last time I checked react-hot-loader was being referenced in the entry and in the `.babelrc` file, not in the module.loaders/rules section. You can check [this sample](https://github.com/fable-compiler/hmr-sample/blob/master/webpack.config.js) for reference which is based in latest react-hot-boilerplate (which uses react-hot-loader 3, still in beta). 

BTW, I need to test if react-hot-loader can be used with elmish. Apparently, besides the configuration you need to put your components in a file different from the entry file and include [some special code](https://github.com/fable-compiler/hmr-sample/blob/8a70d890f70946b530f2fe44b11a697cbe5e566b/src/main.fsx#L20-L23), that I'd like to make more subtle (probably with inlining):

et1975: AFAIK, it reloads (albeit not preserving the state), one just has to open http://localhost:8080/webpack-dev-server/index.html instead of localhost:8080.

PR #43, opened 2017-03-11
Updated counter-ws Readme and renamed variable.

YoLieR:
Thanks for adding the WebSocket example. I updated it's readme to include the necessary install commands and renamed two variables to be consistent.


et1975:
Thanks.

Issue #41, opened 2017-03-08
Try using `requestAnimationFrame` to schedule view renderings instead of calling it directly from the dispatch


et1975:
From evancs' performance harness:

`requestAnimationFrame` lets you say &ldquo;here is a function that modifies the DOM, but I want the browser to make the changes whenever *it* thinks it is a good idea.&rdquo; That means the repaints get smoothed out to 60 FPS no matter how crazy your JS happens to be.

Say you get events coming in extremely quickly, and you get four events within a single frame. A naive use of `requestAnimationFrame` would just schedule them all to happen in sequence. So you would build all four virtual DOMs, diff them against each other in sequence, and show the final result to the user. We can do better though! We can just skip three of those frames entirely. Instead diff the current virtual DOM against the latest virtual DOM. The end result is exactly the same (the user sees the final result) but we skipped 75% of the work!

Elm does this optimization by default.

et1975:
Helps with high rate of change scenarios (like sockets data feed). Committed for react. 

![screen shot 2017-03-09 at 1 13 58 pm](https://cloud.githubusercontent.com/assets/623703/23765634/cce7d5fe-04cf-11e7-8db4-01ee8d8716cb.png)

~8% improvement, but still pretty lacklustre.

MangelMaxime:
Looking at your code, we will provide optimized renderer by default so ? Make sense as the user don't need to understand it. He just want the thing to work :)

Also, it should be easy to port for Elmish-Snabbdom. I will make a PR tomorrow for it.

et1975:
Released in fable-elmish-react@0.8.3

et1975:
@MangelMaxime I'm keeping this open, PR the snabdoom support whenever, but let's keep this open until I figure if it makes sense/difference in ReactNative.

et1975:
Native impl uses React's `setState` which already batches updates, closing.

PR #44, opened 2017-03-13
Elmish-Snabbdom: Optimizing rendering #41

MangelMaxime:
I ported the code from Elmish-React for optimisation using requestAnimationFrame in Elmish-Snabbdom.

When merged we can publish the new version on npm.


et1975:
It may be that snabbdom already has this optimization: https://github.com/snabbdom/snabbdom/issues/32

MangelMaxime:
Ah yes seems like you are right.

No need to do it on our side so.

PR #46, opened 2017-03-15
[WIP] Update to Fable 1.0

alfonsogarciacaro:
Please don't merge it yet as this is a WIP. I still need to write some documentation about the new workflow but you can have an idea with this PR. I already updated the react/counter example so you can check out the branch and try it out (see the updated README for the sample).

As you can see, with Fable 1.0 libraries will be distributing sources directly so there's no need to build anything now (though something will probably be needed for version management). Sorry for changing the name of the folders but, as now we're using project references, it's better to keep the same structure that will be replicated in `node_modules`.

Another thing to note for library authors is the `$(NodeModulesDir)` MSBuild property in Fable.Elmish.fsproj. This is because during development, npm dependencies (like fable-powerpack) will be in the `node_modules` folder, but later they will be on the same level.

Please have a look and tell me if you have any suggestion. I'm planning to announce the beta of Fable 1.0 soon and I'd like to have fable-elmish (with both react and snabbdom renderers) also available.


samples/react/counter/app.fs
```diff
@@ -30,18 +25,18 @@ let update (msg:Msg) count =
...
 let view count dispatch =
   let onClick msg =
     OnClick <| fun _ -> msg |> dispatch
 
   R.div []
-    [ R.button [ onClick Decrement ] [ unbox "-" ]
-      R.div [] [ unbox (string count) ]
-      R.button [ onClick Increment ] [ unbox "+" ] ]
+    [ R.button [ onClick Decrement ] [ R.str "-" ]
+      R.div [] [ R.str (string count) ]
```
et1975: `unbox` was bugging me something fierce, glad to see it go, although why not `text` (like in Elm)?

src/fable-elmish/README.md
```diff
@@ -9,12 +9,12 @@ Elmish abstractions have been carefully designed to resemble Elm's "look and fee
 
 ## Installation
 
+In a project directory with `dotnet-fable` CLI tools installed, type:
+
 ```shell
-npm install --save fable-core fable-powerpack fable-elmish
+dotnet fable add fable-powerpack fable-elmish
```
et1975: Any pointers how this `donet fable` stuff works under the hood?

et1975:
>  distributing sources directly

I for one am happy to see this development :)

Would love to see the FAKE build working, but otherwise it looks so much easier to grasp!

samples/react/counter/app.fs
alfonsogarciacaro: There's already a [`text` function for SVG](https://github.com/fable-compiler/fable-react/blob/6b611f6e196153c7db96f2335e6266fb3f9fbe72/src/react/Fable.Helpers.React.fs#L866) :/

src/fable-elmish/README.md
alfonsogarciacaro: Hehe, actually the `add` command is not working yet. It'll just invoke `npm install fable-powerpack`, then locate the package.json file (in the current or an upper directory) and use the naming convention to add a project reference using the new dotnet sdk command: `dotnet add reference ./node_modules/fable-powerpack/fable.powerpack.fsproj`.

About `dotnet fable` in general, if you have a nuget package named `dotnet-whatever`, you can add it to the project using a `DotNetCliToolReference` tag, the dotnet SDK lets you run the library using commands with the pattern `dotnet whatever`.

alfonsogarciacaro:
@et1975 Glad you like it! Yes, I hope distributing sources solves the problem of users wanting to compile to JS using different Babel plugins, modules, etc. Could you please check the build script yourself? That'd be very helpful! Also, I'd like to publish a pre-release elmish version with this update (using the `next` tag) to check the references work properly when downloading the library through npm, which version number would you like to use?

src/fable-elmish/README.md
et1975: I hope we can use `yarn` instead, at least have an option to.

et1975:
Yep, i can take a look at the build.

alfonsogarciacaro:
Awesome, thank you! So, which version would you like to use for the prerelease? `1.0.0-beta-1`? Or do you want to still keep it below 1.0, e.g `0.9.0-beta-1`?

et1975:
yeah, 0.9.x-yyyy I think. I'm going to move browser-specific (nav) stuff out of the core, as part of the reorg.

et1975:
I'm going to revert the src/folder renames... tried it, not a fan... I'm working in command line mostly and I don't want to have to prefix everything with fable-. We all know it's fable, and the package id has it inside, so don't see a point.

alfonsogarciacaro:
@et1975 The problem is that if the project reference `../fable-elmish/Fable.Elmish.fsproj` from Fable.Elmish.React changes, it won't work when the final user downloads both libraries through npm. You can use MSBuild properties, but Fable right now is doing XML parsing to read the project files and the only property it understands is `$(NodeModulesDir)`.

The only alternative I can think of at the moment is if `fable-elmish-react` references `fable-elmish` also as a an npm package instead of the development sources.

et1975:
>fable-elmish-react references fable-elmish also as an npm package instead of the development sources

Maybe it's a sign that mono-repo was a bad idea to begin with :)

In any case, the current way to import references wouldn't have worked: `node_modules`, which I'd need for PowerPack and React bindings is actually two dirs up `../../node_modules`, which would clash with peer project references using  `..`, wouldn't it? 

How is this?
```
  <ItemGroup>
    <ProjectReference Include="$(NodeModulesDir)/fable-react/Fable.React.fsproj" />
    <ProjectReference Include="$(NodeModulesDir)/fable-elmish/Fable.Elmish.fsproj" Condition="Exists('$(NodeModulesDir)/fable-elmish/Fable.Elmish.fsproj'")/>
    <ProjectReference Include="../elmish/Fable.Elmish.fsproj" Condition="Exists('../elmish/Fable.Elmish.fsproj'")/>
  </ItemGroup>
```

et1975:
Pushed slightly different way to detect and reference peers.

alfonsogarciacaro:
That's the thing, the XML parsing Fable 1.0 is doing atm is very simplistic and doesn't do any advanced MSBuild stuff like conditional items (I just [harcoded `$(NodeModulesDir)` behaviour manually](https://github.com/fable-compiler/Fable/blob/d8c55a15a57de2eda5098eac5349946bb07d83d1/src/dotnet/Fable.Tools/ProjectCracker.fs#L163-L166)). I could use the new library @enricosada created to read the MSBuild args, but I'm not sure yet how well it'll play with the current system to merge all projects and if it will take long to get the arguments for all referenced projects.

Another solution I'm thinking to give more freedom for library developers is to make it a convention to write project references in a Fable library as follow:

```xml
  <!-- REFERENCES DURING DEVELOPMENT -->
  <ItemGroup Condition="!$(MSBuildThisFileDirectory.Contains('node_modules'))" >
    <ProjectReference Include="../../node_modules/fable-react/Fable.React.fsproj" />
    <ProjectReference Include="../elmish/Fable.Elmish.fsproj" />
  </ItemGroup>
  <!-- REFERENCES WHEN USED AS A LIBRARY -->
  <ItemGroup Condition="$(MSBuildThisFileDirectory.Contains('node_modules'))">
    <ProjectReference Include="../../node_modules/fable-react/Fable.React.fsproj" />
    <ProjectReference Include="../elmish/Fable.Elmish.fsproj" />
  </ItemGroup>
```

Again, this would have to be hard coded in Fable.

et1975:
I seem to need (and like having) the local `node_modules` even in development mode :/

alfonsogarciacaro:
Yeah, with that solution you can have `node_modules` for your dependencies in development wherever you want, the key is that during development **Fable.Elmish.React.fsproj itself** won't be contained within `node_modules` and thus `$(MSBuildThisFileDirectory.Contains('node_modules'))` evalutates to false.

My sample above contained some issues so I edited it and added some comments, please tell me if you still have questions.

et1975:
Clearly i don't know what I'm doing. If see a way to get it working I'm happy.

build.fsx
```diff
@@ -35,11 +33,72 @@ let fables  =
...
+let dotnetSDKPath = 
+    System.Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) </> "dotnetcore" 
+    |> FullName
+
+let dotnetExePath =
+    dotnetSDKPath </> (if isWindows then "dotnet.exe" else "dotnet")
+    |> FullName
+
+
+Target "InstallDotNetCore" (fun _ ->
```
forki: we should really submit this as a Task to FAKE

build.fsx
```diff
@@ -74,12 +133,12 @@ Target "Build" (fun _ ->
     |> Seq.iter (fun s -> 
                     let dir = IO.Path.GetDirectoryName s
                     printf "Building: %s\n" dir
-                    Npm (fun p ->
-                        { p with
-                            NpmFilePath = yarn
-                            Command = Run "build"
-                            WorkingDirectory = dir
-                        }))
+                    let result =
```
forki: I think there is a dotnet build task in FAKE

samples/react/counter/app.fsproj
```diff
@@ -0,0 +1,18 @@
...
+  <ItemGroup>
+    <ProjectReference Include="../../../src/fable-elmish-react/Fable.Elmish.React.fsproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="app.fs" />
+  </ItemGroup>
+  <ItemGroup>
+    <DotNetCliToolReference Include="dotnet-compile-fsc" Version="1.0.0-preview2-*" />
+    <PackageReference Include="FSharp.NET.Sdk" Version="1.0.0-beta-060000" PrivateAssets="All" />
+    <PackageReference Include="FSharp.Core" Version="4.1.*" />
```
forki: should be drawn from paket.references ;-)

samples/react/counter/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "build": "node ../node_modules/fable-compiler",
-    "watch": "node ../node_modules/fable-compiler -w",
+    "build": "node ../node_modules/webpack/bin/webpack.js",
+    "watch": "node ../node_modules/webpack/bin/webpack.js -w",
```
forki: this is beautiful ;-)

src/elmish-debugger/Fable.Elmish.Debugger.fsproj
```diff
@@ -1,53 +1,23 @@
...
-    </Otherwise>
-  </Choose>
-  <Import Project="$(FSharpTargetsPath)" Condition="Exists('$(FSharpTargetsPath)')" />
-</Project>
\ No newline at end of file
+  <ItemGroup>
+    <DotNetCliToolReference Include="dotnet-compile-fsc" Version="1.0.0-preview2-*" />
+    <PackageReference Include="FSharp.NET.Sdk" Version="1.0.0-beta-060000" PrivateAssets="All" />
+    <PackageReference Include="FSharp.Core" Version="4.1.*" />
+    <PackageReference Include="Fable.Core" Version="1.0.0-narumi-*" />
```
forki: please put it in references file

samples/react/counter/app.fsproj
```diff
@@ -3,7 +3,7 @@
     <TargetFramework>netstandard1.6</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="../node_modules/fable-elmish-react/Fable.Elmish.React.fsproj" />
+    <ProjectReference Include="../../../src/elmish-react/Fable.Elmish.React.fsproj" />
```
et1975: I like samples using npm dependencies - tells me that the stuff I published works.

src/elmish-debugger/Fable.Elmish.Debugger.fsproj
et1975: @forki can you please PR this? I haven't done much with `dotnet` yet...

forki: I can try when @alfonsogarciacaro is done and everything works

alfonsogarciacaro:
@et1975 Ok, this PR should be more or less to good to go, but I think it's better not to merge it until Fable 1.0 (non-beta) is released so you can still maintain the packages for Fable 0.7.

elmish-debugger and most of the samples are not updated yet, as well as other targets in the FAKE script besides `Build`. It would be tremendously helpful if you could take care of that.

I've published beta versions (with `--tag next`) of fable-elmish and fable-elmish-react, I couldn't do it for fable-elmish-snabbdom because I don't have rights. Please remember the packages should have no dependencies or at least they should be `peerDependencies` to prevent npm downloading different versions of the same Fable package.

@MangelMaxime I've updated the Snabbdom helper API according to what we talked yesterday and a couple of other things that I hope make the code cleaner but I'd like to hear your opinion. As now functions are used to create `VNode`s I turned them to lower case, but we can keep the upper case if you think that's clearer. I've updated the `todomvc` sample, please have a look (you need to publish the fable-elmish-snabbdom package first). There just seem to be some problems with the "active" filter, could you please check?

@enricosada has sent a PR to Ionide to make it compatible with latest FSharp.NET.Sdk so as soon as this is merged and published we can update the `fsproj` files (for example,  `dotnet-compile-fsc` will be gone). Then we can also use Paket to manage nuget dependencies in the `src` projects as @forki suggested (for the samples we could keep the `PackageReference`s for simplicity).

Please ask me if you have any question.

enricosada:
> Then we can also use Paket to manage nuget dependencies in the src projects as @forki suggested (for the samples we could keep the PackageReferences for simplicity

as a note, paket doesnt yet support `dotnet pack` (ref https://github.com/fsprojects/Paket/issues/2191) if that is needed for dev flow

alfonsogarciacaro:
Thanks for pointing that out, @enricosada. fable-elmish and friends will be distributed by source through npm so I guess that should be fine :)

et1975:
@alfonsogarciacaro is the story of transitive dependencies in fable 1.x outlined somewhere?

et1975:
@alfonsogarciacaro trying to import `fable-elmish-debugger` in the samples, why do I get
>Duplicate 'Compile' items were included. The .NET SDK includes 'Compile' items from your project directory by default. You can either remove these items from your project file, or set the 'EnableDefaultCompileItems' property to 'false'

I see you've removed the `EnableDefaultCompileItems` I added and the `src` projects compile w/o it... What am I missing?

et1975:
@alfonsogarciacaro what's the story with ReactNative?

alfonsogarciacaro:
@et1975 About transitive dependencies, basically it's the same situation we have now with Fable 0.7:

- It's not possible (at least it shouldn't) to have references to multiples versions of the same project. If the final app and fable-elmish have a reference to fable-powerpack, they both must use the same version, same as with fable-core. In fact, one of the reasons of not distributing precompiled libraries any more it's to prevent that the libraries were compiled with a different version of fable-core as it happened in the past.

- In terms of .NET the main difference with distributing assemblies is now libraries must keep correct path references to other libraries. That's why we need the convention explained above: libraries should all be at the top level of the `node_modules` folder so references to other libraries from a library (not during the development of the library) should have the form: `../<npm-id>/<project-name>.fsproj`.

I hope this makes sense, we can also think of a way to enforce these conventions later if needed.

About ReactNative, I haven't tested it yet but I guess there should be no problem.

PR #48, opened 2017-03-24
[next] update yarn.lock and misc package.json updates

johlrich:
Some packages were out of sync. Once I updated those and fixed a few others I was able to run the basic samples and had full intellisense with ionide.


et1975:
Thanks for the PR, but we're still working on `next` in pull#46.

PR #49, opened 2017-03-25
Fix #47

### LINKED ISSUE #47: [fable-elmish@next + fable-elmish-snabbdom@next] problems with Fable.Import and Result
A few things I encountered when trying to get running on Fable 1.0:

- `node_modules/fable-elmish-snabbdom/Fable.Import.Snabbdom.fs(L1,13) : error FSHARP: A namespace and a module named 'Fable.Import' both occur in two parts of this assembly`

I think this is caused by the use of `module Fable.Import` inside `Fable.Import.Snabbdom.fs` instead of using `namespace Fable.Import`.

- the `Result` type from `result.fs` inside Elmish conflicts with F#'s `Result`

MangelMaxime:


et1975:
Oops, I made a mistake of merging next into master...

Having said that, 
- I'm removing `Result` from Elmish, it's now available F# 4.1 and in Fable-PowerPack.  
- and moving the browser-specific stuff (i.e. navigation) into a separate package.

F# 4.1 is also a prerequisite going forward and it does not require `CompilationRepresentationFlags.ModuleSuffix`, since it is now a default behaviour of the compiler.

Closing.

MangelMaxime:
Yes I was wondering why next was merged. :)

Ok about Result etc. I will wait for the next version of the navigation package before finishing the docs site.

PR #58, opened 2017-03-29
update debugger serialization for fable v1

johlrich:
* use `deflate` with remotedev's `serialize.reducer` option
* workaround for init serialization
* use new reflection apis for DU name

### DIFF
--- a/src/elmish-debugger/debugger.fs
+++ b/src/elmish-debugger/debugger.fs
@@ -5,30 +5,45 @@ open Fable.Core
 
 [<RequireQualifiedAccess>]
 module Debugger =
+    open FSharp.Reflection
+
+    let inline private duName (x:'a) = 
+        match FSharpValue.GetUnionFields(x, typeof<'a>) with
+        | case, _ -> case.Name
+
+    let [<PassGenericsAttribute>] private getCase cmd : obj =
+        createObj ["type" ==> duName cmd
+                   "msg" ==> cmd]
+
     type ConnectionOptions =
         | ViaExtension
         | Remote of address:string * port:int
         | Secure of address:string * port:int
 
     let connect =
-        let inline getCase cmd : obj = createObj ["type" ==> unbox cmd?tag
-                                                  "fields" ==> cmd?fields]
+        let serialize = createObj ["replacer" ==> fun _ v -> deflate v]
 
         function
-        | ViaExtension -> { Options.remote = false; hostname = "localhost"; port = 8000; secure = false; getActionType = Some getCase }
-        | Remote (address,port) -> { Options.remote = true; hostname = address; port = port; secure = false; getActionType = None }
-        | Secure (address,port) -> { Options.remote = true; hostname = address; port = port; secure = true; getActionType = None }
+        | ViaExtension -> { Options.remote = false; hostname = "localhost"; port = 8000; secure = false; getActionType = Some getCase; serialize = serialize }
+        | Remote (address,port) -> { Options.remote = true; hostname = address; port = port; secure = false; getActionType = None; serialize = serialize }
+        | Secure (address,port) -> { Options.remote = true; hostname = address; port = port; secure = true; getActionType = None; serialize = serialize }
         >> connectViaExtension
 
 [<RequireQualifiedAccess>]
 module Program =
     open Elmish
 
+    [<Emit("JSON.parse($0)")>]
+    let private parse str : obj = jsNative
+
     [<PassGenericsAttribute>]
     let withDebuggerUsing (connection:Connection) (program : Program<'a,'model,'msg,'view>) : Program<'a,'model,'msg,'view> =
         let init a =
             let (model,cmd) = program.init a
-            connection.init (model, None)
+            // simple looking one liner to do a recursive deflate
+            // needed otherwise extension gets F# obj
+            let decoded = model |> toJson |> parse
+            connection.init (decoded, None)
             model,cmd
 
         let update msg model : 'model * Cmd<'msg> =
--- a/src/elmish-debugger/debugger.fs
+++ b/src/elmish-debugger/debugger.fs
@@ -11,7 +11,7 @@ module Debugger =
         match FSharpValue.GetUnionFields(x, typeof<'a>) with
         | case, _ -> case.Name
 
-    let [<PassGenericsAttribute>] private getCase cmd : obj =
+    let [<PassGenericsAttribute>] inline private getCase cmd : obj =
         createObj ["type" ==> duName cmd
                    "msg" ==> cmd]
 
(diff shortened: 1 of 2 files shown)


src/elmish-debugger/debugger.fs
```diff
@@ -5,30 +5,45 @@ open Fable.Core
...
+    let private parse str : obj = jsNative
+
     [<PassGenericsAttribute>]
     let withDebuggerUsing (connection:Connection) (program : Program<'a,'model,'msg,'view>) : Program<'a,'model,'msg,'view> =
         let init a =
             let (model,cmd) = program.init a
-            connection.init (model, None)
+            // simple looking one liner to do a recursive deflate
+            // needed otherwise extension gets F# obj
+            let decoded = model |> toJson |> parse
```
johlrich: Went with this weird toJson / parse combination so I didn't have to write a method that applied deflate recursively on the JS object, which would be a more efficient way to go about this.

src/elmish-debugger/debugger.fs
```diff
@@ -5,30 +5,45 @@ open Fable.Core
...
+
     type ConnectionOptions =
         | ViaExtension
         | Remote of address:string * port:int
         | Secure of address:string * port:int
 
     let connect =
-        let inline getCase cmd : obj = createObj ["type" ==> unbox cmd?tag
-                                                  "fields" ==> cmd?fields]
+        let serialize = createObj ["replacer" ==> fun _ v -> deflate v]
```
johlrich: I originally tried giving the replacer and reviver fns to this option, but I couldn't get the extractState working when I did that, so I ended up leaving the extract state management alone and just  supplied the replacer.

johlrich:
I wondered if it made sense to try and support actions that wouldn't be DUs as well? Something like check if union, if so get case, otherwise call something like `sprintf "%A"`

Or is that such a common practice that we shouldn't worry about it?

johlrich:
Oh one last thing, I couldn't use the extension to jump state. I was able to debug and see the properly inflated state was being passed to `program.setState`, so I don't think these changes are why,  but sure 100% either.

et1975:
I realize I haven't provided [the context](https://github.com/zalmoxisus/remotedev/issues/6) for intended serialization changes - TD;LR: `inflate`/`deflate` are crutches that may or may not work on a given type and may or may not remain a public Fable feature, consequently the full `toJson`/`ofJson` is the preferred/foolproof way to get the roundtrip serialization going. The fancier option would be to implement replacer/reviver and do the serialization *per field*, a la ImmutableJS, possibly with a nice looking [preview](https://cloud.githubusercontent.com/assets/7957859/21814330/a17d556a-d761-11e6-85ef-159dd12f36c5.png).

src/elmish-debugger/debugger.fs
```diff
@@ -5,30 +5,45 @@ open Fable.Core
 
 [<RequireQualifiedAccess>]
 module Debugger =
+    open FSharp.Reflection
+
+    let inline private duName (x:'a) = 
+        match FSharpValue.GetUnionFields(x, typeof<'a>) with
+        | case, _ -> case.Name
+
+    let [<PassGenericsAttribute>] private getCase cmd : obj =
```
et1975: this could be another `inline`

src/elmish-debugger/debugger.fs
```diff
@@ -5,30 +5,45 @@ open Fable.Core
...
     let connect =
-        let inline getCase cmd : obj = createObj ["type" ==> unbox cmd?tag
-                                                  "fields" ==> cmd?fields]
+        let serialize = createObj ["replacer" ==> fun _ v -> deflate v]
 
         function
-        | ViaExtension -> { Options.remote = false; hostname = "localhost"; port = 8000; secure = false; getActionType = Some getCase }
-        | Remote (address,port) -> { Options.remote = true; hostname = address; port = port; secure = false; getActionType = None }
-        | Secure (address,port) -> { Options.remote = true; hostname = address; port = port; secure = true; getActionType = None }
+        | ViaExtension -> { Options.remote = false; hostname = "localhost"; port = 8000; secure = false; getActionType = Some getCase; serialize = serialize }
```
et1975: These record constructors are now big enough to warrant a `fallback` record above (with defaults) and per-case `{ fallback with }` to update just the relevant fields.

johlrich: No problem. Two questions:

1) Should `getActionType` still be None for `Remote` and `Secure` or should all of them use `getCase`? I've only used this module with the extension so wasn't sure for the reason it wasn't used in the past.
2) See any problem with using the remotedev defaults for fallback where there is overlap in settings? https://github.com/zalmoxisus/remotedev/blob/master/src/constants.js#L2-L4

et1975: `getActionType` used to be implemented only in the extension, Mihail mentioned it will be unified in v3 and work for remote/native debugging as well, not sure if it that's the case.

I'm ambivalent about the defaults, whatever makes sense 👍

src/elmish-debugger/debugger.fs
```diff
@@ -5,30 +5,47 @@ open Fable.Core
 
 [<RequireQualifiedAccess>]
 module Debugger =
+    open FSharp.Reflection
+
+    let inline private duName (x:'a) = 
+        match FSharpValue.GetUnionFields(x, typeof<'a>) with
+        | case, _ -> case.Name
+
+    let [<PassGenericsAttribute>] inline private getCase cmd : obj =
```
et1975: ` [<PassGenericsAttribute>]` is redundant once the function is `inline`d

johlrich: Oops, I guess I don't fully understand `PassGenericsAttribute` yet. What makes its need go away once inlined? I had assumed it was needed anytime there was a `typeof`. Changing now.

johlrich:
> `inflate`/`deflate` are crutches that may or may not work on a given type and may or may not remain a public Fable feature, consequently the full `toJson`/`ofJson` is the preferred/foolproof way to get the roundtrip serialization going

Since `toJson`/`ofJson` use `deflate`/`inflate` to do their job, I suppose I don't see them as being that far off the preferred way. As `toJson`/`ofJson` support more types, we will get that support with no changes.

Still, I do agree that having something more tailored for the dev-tooling with the nice previews, etc would be best long term.

`deflate` in particular is already a `replacer` with no `key` argument, so could be a good place to start to look into that next step

src/elmish-debugger/debugger.fs
et1975: inlined functions are like compiler-approved macros, the body (code that *uses* generic information) is embedded at the callsite. At that point it's up to that method to have the type info, if the type is locally determined - all is well, if the type comes from elsewhere we recurse (attribute/inline), until we're all the way up the call chain.

et1975:
> Since toJson/ofJson use deflate/inflate to do their job, I suppose I don't see them as being that far off

I'm basing my guidance on an earlier discussion we had with Alfonso on the topic, so let's appeal to authority... accio @alfonsogarciacaro !

Alfonso, what can we expect of the current and future inflate/deflate vs toJson/ofJson duality?

src/elmish-debugger/debugger.fs
johlrich: Thanks for that explanation! Makes perfect sense now.

alfonsogarciacaro:
@johlrich @et1975 If `inflate/deflate` are working for you, I don't see any problem in tying them to `ofJson/toJson` (actually I don't see any reason not to). I was a bit cautious to make them public only because users could think they apply to anything, for example a user was trying to inflate a DU serialized as default by Json.NET. But if you inflate and object as has been deflated by `deflate` (in the case of DUs, following the [FSharpLu.Json](https://github.com/Microsoft/fsharplu/wiki/fsharplu.json) format) there shouldn't be any problem.

We can leave a comment in Serialize.ts as reminder so I or future maintainers don't forget :)

et1975:
Ok, with serialization out of the way, @johlrich you are saying going to a historical state does not work... is the debugger a last `program` modifier in the chain? It has to be, cause the renderer would provide `setState` that the debugger would need.

johlrich:
@et1975 Ahhh that was it, I had it in the wrong order on the application I was testing. I double checked my pre-beta app and it also had the debugger call at the end. I must have swapped them at some point when porting. Thanks.

et1975:
Thanks!

Issue #50, opened 2017-03-26
Create templates for fable-elmish-react and fable-elmish-react-native projects


alfonsogarciacaro:
@et1975 @MangelMaxime Could you please create templates for fable-elmish-react and fable-elmish-snabbdom projects? You can take the [simple Fable template](https://github.com/fable-compiler/Fable/tree/narumi/src/templates/simple) as reference. Or if you prefer, you can just add the files and I'll take care of the .nuspec and template.json configuration 👍 

> The [documentation for dotnet SDK templates](https://github.com/dotnet/templating/wiki/%22Runnable-Project%22-Templates) is still incomplete but there's [a tutorial here](http://pioneercode.com/post/how-to-create-a-dot-net-new-project-template-in-dot-net-core) if you need help.

alfonsogarciacaro:
Now that https://github.com/fable-compiler/Fable/issues/781 it'd be nice to have a template ready for F# eXchange, so basically I'd need it... NOW! sorry ;)

A more or less full-app with navigation and debugger will be nice. You can just do the code and I'll take care of the template config and Nuget publishing. The only thing you need to remember is you can use a string like `FableElmishReactTemplate` (no periods or hyphens allowed, I think) that will be replaced **both in file name and content** with the project name given by the user.

MangelMaxime:
@alfonsogarciacaro Nice news :)

I was waiting for it to work again with Elmish. I should have the time to create the template today or tomorrow.

et1975:
Blocked, waiting for https://github.com/fable-compiler/Fable/issues/804

et1975:
Closing, will be tracked in [templates](fable-elmish/templates) repo if needed

PR #59, opened 2017-04-03
Template for Elmish + React

MangelMaxime:
**Please do not merge yet this PR**

This PR is related  #50.

The template include:
- Elmish
- Navigation support
- Counter sample used as a nested app
- Remote debugger support (tested under firefox addon)
- SASS compilation + bundle via Webpack
- .editorconfig file to ensure same indentation, etc. across the editors
- yarn support via a `yarn.lock` file (optional users can still use npm... :smile:  )

I didn't tested the production generation of Webpack.
Would be nice to include Fable-Aether (the project need to be updated for Fable 1.0 beta). From previous project, using aether help a lot having cleaner nested application support.

@alfonsogarciacaro I didn't succeed when trying to install the template from the file system. Via `dotnet new --install path/to/template` so I didn't create the nuget.

Can you please take a look at the template. See if it's suit your needs and if the code is clean enough.
If all is ok can you please create the nuget etc. ? We will update it with the next features later it's not a problem on our side.


alfonsogarciacaro:
Awesome work, @MangelMaxime, thank you! :clap: :clap:

Sorry, I should have explained you better. I think you need to create the package first. Unfortunately using `nuget.exe` (dotnet pack cannot read .nuspec files yet apparently), download it from [here](https://dist.nuget.org/index.html) and type `mono path/to/nuget.exe Fable.Template.nuspec`, then you can `dotnet new --install path/to/Fable.Template.0.1.0.nupkg`.

Anyways, I'll take care of the publishing :+1:

What do you need to update in Aether? Is it not working with Fable 1.0 just by adding the Aether.fs file?

MangelMaxime:
From my quick test the ReduxDevTools is working with this application.
Even the history navigatoin. I just got some glitch on the input of the Home page.

I need aether to clean this code:

```fs
    | CounterMsg msg ->
        let (res, _) = Counter.update msg model.SubModels.Counter
        let localUpdate counter = { model.SubModels with Counter = counter }
        { model with SubModels = localUpdate res }, Cmd.Empty
```

into something like

```fs
    | CounterMsg msg ->
        let (res, _) = Counter.update msg model.SubModels.Counter
        Optic.set (Model.SubModels >-> SubModels.Counter) counter model, Cmd.Empty
```

This come really needed when nesting a lot of application as updating nested record is painful by default in F# :).
We could include directly the Aether.fs file I used in Fable-Arch. But would be nice to use Fable-Aether directly :)

MangelMaxime:
One thing I am thinking about.
When checking the template for generating the production files, we could also add a `#if DEBUG` around the debugger features.

This will avoid bundling the Debugger features in the production bundle.

alfonsogarciacaro:
Ups, I think I forgot to click the Submit review button until now. Were my comments visible?

@MangelMaxime I think using `#if DEBUG` is a very good idea. Another trick is to use the `[<Conditional("DEBUG")>]` attribute on the methods you want to be erased in production mode :)

MangelMaxime:
@alfonsogarciacaro Yes your comments are visible :)

I will send an update taking them in account :)

src/templates/react/Content/.template.config/template.json
```diff
@@ -0,0 +1,15 @@
...
+  "author": "Maxime Mangel",
+  "classifications": [ "Fable" ],
+  "name": "Simple elmish application using React as a renderer",
+  "tags": {
+    "language": "F#",
+    "type": "project"
+  },
+  "identity": "Fable.Template.Elmish.React",
+  "groupIdentity": "Fable.Elmish",
+  "shortName": "FableElmishReact",
```
MangelMaxime: Personally I think using `fable-elmish-react` make sense because we will also have `fable-elmish-react-native`.

@et1975 What is your opinion ?

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,152 @@
...
+open Fable.Import
+open Fable.Import.Browser
+open FableElmishReactTemplate.Common
+
+module App =
+
+  [<Emit("require('../sass/main.sass');")>]
+  let requireStyle () = ()
+
+  requireStyle()
```
MangelMaxime: I was thinking react need `require` keyword but this seems to work too :)
And it's cleaner.

src/templates/react/Content/webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+    contentBase: resolve('./public'),
+    port: 8080
+  },
+  module: {
+    rules: [
+      {
+        test: /\.fs(x|proj)?$/,
+        use: {
+          loader: "fable-loader",
+          options: { babel: babelOptions }
```
MangelMaxime: How do you set your env to be "production" ?
Seems like doing `node_modules/.bin/webpack-dev-server -p` isn't enough.

I am using `dotnet fable start` that's why I'm using webpack directly.

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+  open Fable.Helpers.React.Props
+
+  let menuItem label page currentPage dispatch =
+      li
+        [ ]
+        [ a
+            [ classList [ "is-active", page = currentPage ]
+              OnClick (fun _ -> NavigateTo page |> dispatch)
+            ]
+            [ unbox label ]
```
alfonsogarciacaro: Sorry I forgot to say it before, could you please change this and other `unbox` to `str` (from Fable.React.Helpers)?

src/templates/react/Content/.template.config/template.json
```diff
@@ -0,0 +1,15 @@
+{
+  "author": "Maxime Mangel",
+  "classifications": [ "Fable" ],
+  "name": "Simple elmish application using React as a renderer",
```
alfonsogarciacaro: This looks too long on screen, could you please change it to `Fable.Elmish.React App`? Thanks!
screen shot 2017-04-05 at 17 14 39

alfonsogarciacaro:
@MangelMaxime I gave it a try and it's working, including the Redux dev tools. Really cool, thank you! I just left a couple of additional comments. Unfortunately, I had to manually copy latest FCS dll to make it work with Ionide, but I'll try to convince @Krzysztof-Cieslak tomorrow to publish a new Ionide release ;)

I published it to nuget (install it with `dotnet new -i Fable.Template.Elmish.React::*`). I want to make you an owner too, do you have a nuget username Maxime?

src/templates/react/Content/public/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html>
+<head>
+  <meta http-equiv='Content-Type' content='text/html; charset=utf-8'>
+  <title>Elmish  Fable App</title>
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css" />
+  <script src="https://cdn.polyfill.io/v2/polyfill.js?features=default,fetch"></script>
```
et1975: Babel runtime should have provided all the polyfills, no?

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+open Fable.Import
+open Fable.Import.Browser
+open FableElmishReactTemplate.Common
+
+module App =
+
+  importAll "../sass/main.sass"
+
+  type SubModels =
+    {
```
et1975: This is not idiomatic placement of {}s

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+      Home: Home.Model
```
et1975: Since interop with .NET is not an issue (the source of CamelCase notation in F#), I think we should use JS (lowerCase) conventions when it comes to names.

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+    | CounterMsg of Counter.Msg
+    | HomeMsg of Home.Msg
+    | NavigateTo of Page
+
+  let update msg model =
+    match msg with
+    | CounterMsg msg ->
+        let (res, _) = Counter.update msg model.SubModels.Counter
+        let localUpdate counter = { model.SubModels with Counter = counter }
+        { model with SubModels = localUpdate res }, Cmd.Empty
```
et1975: `Cmd.Empty` really threw me off! It should be `Cmd.none` or `[]`

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+    | Some page ->
+        { model with CurrentPage = page }, []
+
+  let init result =
+    urlUpdate result Model.Initial
+
+  type Msg =
+    | CounterMsg of Counter.Msg
+    | HomeMsg of Home.Msg
+    | NavigateTo of Page
```
et1975: Page navigation is already handled as part of the navigable app, we should guide the user towards relying on those facilities instead of burdening `update` function with it.

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+
+  open Fable.Helpers.React
+  open Fable.Helpers.React.Props
+
+  let menuItem label page currentPage dispatch =
+      li
+        [ ]
+        [ a
+            [ classList [ "is-active", page = currentPage ]
+              OnClick (fun _ -> NavigateTo page |> dispatch)
```
et1975: IMHO this should be simple local link, like `Link (toHash page)`

src/templates/react/Content/src/App.fs
```diff
@@ -0,0 +1,149 @@
...
+    | NavigateTo of Page
+
+  let update msg model =
+    match msg with
+    | CounterMsg msg ->
+        let (res, _) = Counter.update msg model.SubModels.Counter
+        let localUpdate counter = { model.SubModels with Counter = counter }
+        { model with SubModels = localUpdate res }, Cmd.Empty
+    | HomeMsg msg ->
+        let (res, _) = Home.update msg model.SubModels.Home
```
et1975: child cmd should never be discarded, even as an example - it breaks the encapsulation!

alfonsogarciacaro:
@MangelMaxime Sent you a PR ;) #61

src/templates/react/Content/FableElmishReactTemplate.fsproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.6</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="src/Common.fs" />
+    <Compile Include="src/Navbar.fs" />
+    <Compile Include="src/Pages/About.fs" />
```
et1975: This is highly subjective of course, but personally something [like this](http://blog.jenkster.com/2016/04/how-i-structure-elm-apps.html) is the file organization I encourage.

src/templates/react/Content/src/Navbar.fs
```diff
@@ -0,0 +1,48 @@
...
+          ]
+        span
+          [ ]
+          [ unbox txt ]
+      ]
+
+  let navButtons =
+    span
+      [ ClassName "nav-item"
+      ]
```
et1975: Again, I would "idiomaticise" those []s.

src/templates/react/Content/public/index.html
alfonsogarciacaro: Good point, @et1975, I didn't notice this line. Didn't know about polyfill.io either. Looks very interesting, as it only provides the polyfills needed for the current browser. It also provides `fetch` compatibility, which doesn't come in core-js. Maybe we could just disable core-js in the babel-transform-runtime plugin and use this instead?

et1975: Less is more I say 👍

et1975:
Re: Aether - I think repackaging it for Fable 1.x is nice to have, but it won't add anything substantial, just make `ProjectReference` import possible. Regardless, I wouldn't add it to the template, it can get in the way of understanding what's going on.

src/templates/react/Content/src/App.fs
MangelMaxime: I agree with you :)

src/templates/react/Content/src/Navbar.fs
MangelMaxime: I was using this notation because it's helping viewing the end a the current "block".
As the starting `[` is aligned with the ending `]`.

Anyway, I will try to use the "idiomatic" way for those []s.

src/templates/react/Content/src/App.fs
MangelMaxime: @et1975 I suppose you mean something like : `Href (toHash page)`
I don't know what `Link` stand for in your comment :)

et1975: Yes, `Href`, sorry.

MangelMaxime:
@alfonsogarciacaro I just created a nuget account. Here is my username: MangelMaxime

MangelMaxime:
@et1975 You are probably right about Aether.

We will be able to add the information Aether, good practices etc. over the docs site.
I almost finished to port the code. I will soon work on the sample generation via Fake etc.

MangelMaxime:
@alfonsogarciacaro @et1975 
Ok, so I made a huge refactoring of the template.

I remove the `namespace` usage and use `module` declaration everywhere.
I took inspiration from [link given by Eugene](http://blog.jenkster.com/2016/04/how-i-structure-elm-apps.html) and also the react native navigation sample.

The refactor includes:
- Idiomatic syntax for the `{}` and the `[]`
- Better file splitting
- Use `str` instead of `unbox` has shown by Alfonso
- More simple [`Model`](https://github.com/fable-compiler/fable-elmish/blob/ed8892dacfdd17517b4adfe108f4766a942c9960/src/templates/react/Content/src/Types.fs#L10-L14) type definition for the top App. (Avoid nested record update)
- Lowercase record fields name
- No discard of the Sub application commands [Implementation](https://github.com/fable-compiler/fable-elmish/blob/ed8892dacfdd17517b4adfe108f4766a942c9960/src/templates/react/Content/src/State.fs#L45-L52)

Also, I change the way [`urlUpdate`](https://github.com/fable-compiler/fable-elmish/blob/ed8892dacfdd17517b4adfe108f4766a942c9960/src/templates/react/Content/src/State.fs#L17-L34) is working. So now, we don't maintain state between page switch. I don't know if we want this behavior or old one.
The benefit of the current implementation is to show a more classic way of initializing a sub application (support for model initial value and initial command). This is useful when you need to send an Ajax request for example to fetch initial data for the page.

The fact we use modules now, allow us to have a cleaner implementation. For example, if you are want to access the views component simply type `Views.` and the intellisense will give you all the "view components" available at this level.

I made a module called `Global` which contains the `Page` DU, and `toHash` function as they could be use from sub applications. (redirection, quick home return etc.)

That's should be all the most important changes. Feedbacks always welcome :)

Still to be implemented: 
- DEBUG symbols support ( @alfonsogarciacaro I didn't understand how to switch webpack into production mode. I tried using `-p` but did nothing according to my tests).
- Polyfill support. I didn't understand if you guys said we should use core-js or polyfill.io

Ping @whitetigle as I know you started an Elmish project and perhaps the discussion will interest you.

src/templates/react/Content/src/State.fs
```diff
@@ -0,0 +1,52 @@
...
+
+let urlUpdate (result: Option<Page>) model =
+  match result with
+  | None ->
+    console.error("Error parsing url")
+    ( model,Navigation.modifyUrl (toHash model.currentPage) )
+  | Some page ->
+      match page with
+      | Counter ->
+          let (counter, counterCmd) = State.Counter.init ()
```
et1975: This is interesting... I always do initialization only once, in the `init`!

src/templates/react/Content/src/State.fs
```diff
@@ -0,0 +1,52 @@
...
+    map About (s "about")
+    map Counter (s "counter")
+    map Home (s "home")
+  ]
+
+let urlUpdate (result: Option<Page>) model =
+  match result with
+  | None ->
+    console.error("Error parsing url")
+    ( model,Navigation.modifyUrl (toHash model.currentPage) )
```
et1975: unnecessary outer `()`s

src/templates/react/Content/src/State.fs
```diff
@@ -0,0 +1,52 @@
...
+      | _ -> { model with currentPage = page }, []
+
+let initialModel = {
+    currentPage = Home
+    counter = 0
+    home = ""
+  }
+
+let initApp result =
+  urlUpdate result initialModel
```
et1975: This is where I'd do the child initialization

src/templates/react/Content/src/State.fs
MangelMaxime: @et1975 
How would you handle application which have an initial model state and an initial pool of `Cmd` ?
For example, your sub application need to fetch the list of the users from the server.

The solution could be something like:

```fs
      let initialCmd =
        match page with
        | Counter -> Cmd.map CounterMsg (State.Counter.initCmd ())
        | _ -> []
      { model with currentPage = page }, initialCmd
```

et1975: Here's an example from our codebase:

```fsharp
let init result : Model * Cmd<Msg>  =
    let (auth,authCmd)=
        match result with 
        | Some (Logon jwt) -> Auth.State.init (Some jwt)
        | _ -> Auth.State.init None
    let (ass,assCmd) = Associations.State.init ()
    let (sensors,sensorsCmd) = Sensors.State.init ()
    let (settings,settingsCmd) = Settings.State.init ()
    let model,cmd =
       urlUpdate result
         { page = Home
           authentication = auth
           associations = ass
           sensors = sensors
           settings = settings }
    model, Cmd.batch [ cmd
                       Cmd.map AssociationsMsg assCmd
                       Cmd.map SensorsMsg sensorsCmd
                       Cmd.map SettingsMsg settingsCmd
                       Cmd.map AuthMsg authCmd ]
```

MangelMaxime: Oh ok I see.

The missing part in my mind was:

```fs
 let model,cmd =
       urlUpdate result
         { page = Home
           ...
```

Will update the template this evening with it.

MangelMaxime:
Ok, so the template should be finish now.

We probably need to update the README.md to explan how to use it (architecture, build/dev process, etc.). Will do this during this week.

src/templates/react/Content/src/App.fs
```diff
@@ -66,11 +66,15 @@ let root model dispatch =
                     [ pageHtml model.currentPage ] ] ] ] ]
 
 open Elmish.React
+#if DEBUG
 open Elmish.Debug
```
et1975: @alfonsogarciacaro correct me if I'm wrong, but I think webpack will eliminate dead code, so opening a namespace is harmless unless one uses something from it.

MangelMaxime: Ah yes didn't think about treeshaking here. @et1975 You are probably right.

et1975:
@MangelMaxime looks good, just remove WIP when you think it's ready.

src/templates/react/Content/src/App.fs
alfonsogarciacaro: Not only that, Fable won't include any reference to the file if you don't actually use anything from there (opening is just not enough). It could be argued that this is a bug because you may not get side effects when opening a module, but unfortunately the F# AST doesn't give the open statements, only the actual calls.

Anyways, TL;DR; Yes, it's not necessary to add `#if DEBUG` around open statements.

alfonsogarciacaro:
@MangelMaxime Awesome work, thanks a lot for this! I'm sure this will become an incredible resource for all fable-elmish users.

> DEBUG symbols support ( @alfonsogarciacaro I didn't understand how to switch webpack into production mode. I tried using -p but did nothing according to my tests).

This is tricky, when you pass `-p` to Webpack, NODE_ENV is set to "production" [but not within webpack config](https://webpack.js.org/guides/production-build/#node-environment-variable). The solutions are using `crossenv` in the npm script or just detecting if there exists a `-p` (or `--production`) argument in webpack.config.js.

MangelMaxime:
@alfonsogarciacaro Thanks :)

Yes I saw this was a tricky problem and took inspiration from the fable-suave-scaffold sample.  Making it more simple to write the build script as I have direct access to node_modules.

src/templates/react/Content/src/App.fs
et1975: opening a module is not enough even in good-old F# on .NET: they act exactly as static class, i.e. static constructors and fields (and let bindings) are guaranteed to be initialized "at some point" before the *first use*... no use - no initialization!

MangelMaxime: Ok, I will make the change in the template so :)

PR #66, opened 2017-04-11
Add NuGet.Config for dotnet restore to work on funcy machines

forki:
just working around the quirks of THAT dependency manager


et1975:
So... this just goes in the root? Since restore is called per-project what's using this?

forki:
Nuget is stupid. It looks up the path and then into machine settings and so on. So this just breaks the chain at the root.

PR #67, opened 2017-04-12
Adding a counter list sample

forki:
This sample code shows how to reuse / stack elmish components.

list

### DIFF
--- a/samples/react/counter-list/app.fs
+++ b/samples/react/counter-list/app.fs
@@ -1,0 +1,76 @@
+module App
+
+(**
+ - title: Counter-List
+ - tagline: The famous Increment/Decrement list sample ported from Elm
+*)
+
+open Fable.Core
+open Fable.Import
+open Elmish
+
+// MODEL
+
+type ID = int
+
+type Model = {
+    Counters : Counter.Model list
+}
+
+type Msg = 
+| Insert
+| Remove
+| Modify of ID * Counter.Msg
+
+let init() : Model =
+    { Counters = [] }
+
+// UPDATE
+
+let update (msg:Msg) (model:Model) =
+
+    match msg with
+    | Insert ->
+        let newCounter = Counter.init()
+        { Counters = newCounter :: model.Counters }
+
+    | Remove ->
+        { Counters = List.tail model.Counters }
+
+    | Modify (id, counterMsg) ->
+        { model with
+            Counters =
+                model.Counters
+                |> List.mapi (fun i counterModel -> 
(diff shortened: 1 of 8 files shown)


samples/react/counter-list/counter.fs
```diff
@@ -0,0 +1,43 @@
...
+
+  R.div []
+      [ R.button [ OnClick (fun _ -> dispatch Decrement) ] [ R.str "-" ]
+        R.div [] [ R.str (string model) ]
+        R.button [ OnClick (fun _ -> dispatch Increment) ] [ R.str "+" ] ]
+
+open Elmish.React
+
+// App
+Program.mkSimple init update view
```
et1975: ?!

forki: Could you please put a a question before the question mark ;-)

et1975: why does it define a program, when it's a child component?

forki: Ah. That's the counter.fs - it's just copy paste. I will remove that

forki: done

samples/react/counter-list/app.fs
```diff
@@ -0,0 +1,76 @@
...
+module R = Fable.Helpers.React
+
+// VIEW (rendered with React)
+
+let view model dispatch =
+    let remove = R.button [ OnClick (fun _ -> dispatch Remove) ] [ R.str "Remove" ]
+
+    let insert = R.button [ OnClick (fun _ -> dispatch Insert) ] [ R.str "Add" ]
+
+    let counterDispatch i msg = dispatch (Modify (i, msg))
```
forki: @et1975 any ideas how to make the counter dispatch function nicer? TBH I don't know how to get rid if it. But maybe it's fine?!

et1975: Elm avoids explicit `dispatch` arg in the view by providing all the HTML properties that can result in a message as functions that return one (`fun 'a -> 'msg` instead of `fun 'a -> unit`) .

forki: it's complicated. sometimes subview needs to send application wide messages.

et1975: Child shouldn't know anything about it's place in the hierarchy, it's responsibility of container to inspect child messages if there's a need to propagate the event outside of the child.

forki: ok that sounds like the better way to me. do you have an exampe?

PR #68, opened 2017-04-12
Refactor module names. Avoid namespace collision for nested components

MangelMaxime:
This refactor is needed when working with nested components.

This avoid namespace collision like:

```fs
module Views.Forum
//...

module Views.Forum.Thread 
// Collision here Views.Forum already defined
//...
```

New organization:
```fs
module Forum.View
//...

module Forum.Thread.View 
// Ok this time :)
//...
```

Edit: Sorry for this problem and new *major* change. 
But I didn't encounter this problem before testing on a real case app.

### DIFF
--- a/src/templates/react/Content/FableElmishReactTemplate.fsproj
+++ b/src/templates/react/Content/FableElmishReactTemplate.fsproj
@@ -5,8 +5,8 @@
   <ItemGroup>
     <!-- Global to the app -->
     <Compile Include="src/Global.fs" />
-    <!-- About -->
-    <Compile Include="src/About/View.fs" />
+    <!-- Info -->
+    <Compile Include="src/Info/View.fs" />
     <!-- Counter -->
     <Compile Include="src/Counter/Types.fs" />
     <Compile Include="src/Counter/State.fs" />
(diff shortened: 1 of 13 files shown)


alfonsogarciacaro:
@enricosada We're having issues when we forget to delete `node_modules` before publishing the template, is there a way to exclude it from the package.

Also, I was following the tutorial you sent me (which I cannot find now) to make the template. That tutorial used a .nuspec file and AFAIK that's not compatible with `dotnet pack`. Is there a better way to create the templates? Should we open an issue in dotnet/templating?

enricosada:
you can use `<NuspecFile>` to choose the nuspec to use inside a `.proj` file (or csproj, or fsproj, doenst matter) and will worj with `dotnet pack`. But that's the last solution.

let me check if is possibile to fix that

enricosada:
@alfonsogarciacaro i converted to use a normal msbuild proj file instead of nuspec (that's just an intermediate file now with new sdk), so is compatibile with sdk (`dotnet pack` and props)
And excluded `node_modules` dir

-> https://github.com/fable-compiler/fable-elmish/pull/80

PR #69, opened 2017-04-12
Removed dispatch paramter that is not used

Zaid-Ajaj:


et1975:
Looks good to me, if @MangelMaxime is ok with it?

MangelMaxime:
Ah yes.
It's an old code before we use the `Href (toHash x)`.

Thanks @Zaid-Ajaj

PR #70, opened 2017-04-12
Added calculator sample

Zaid-Ajaj:
Hopefully the structure is in accordance with the convensions here, this time I added handling negative numbers as well. I liked @forki's idea of using gifs for previewing :smile:

elmish-calc

### DIFF
--- a/samples/react/ElmishCalc/src/State.fs
+++ b/samples/react/ElmishCalc/src/State.fs
@@ -1,0 +1,77 @@
+module Calc.State
+
+open Calc.Types
+
+/// Active pattern that matches with an operation
+let (|Operation|_|) = function 
+    | Plus -> Some Plus
+    | Minus -> Some Minus
+    | Times -> Some Times
+    | Div -> Some Div
+    | _ -> None
+
+/// Given a model, calculate the answer
+let solve (state : Input list) = 
+  match state with
+  | [Const x; Operation op; Const y] ->
+      match op with
+      | Plus -> Some (x + y)
+      | Minus -> Some (x - y)
+      | Times -> Some (x * y)
+      | Div when y = 0 -> None // division by zero not allowed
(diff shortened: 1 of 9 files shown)


samples/react/ElmishCalc/src/State.fs
```diff
@@ -0,0 +1,77 @@
+module Calc.State
+
+open Calc.Types
```
MangelMaxime: When opening modules at the "same level", you could simply do `open Types`.
This allow an easier read I think.

And only specify the full namespace when using nested component :)

This comment apply to all `open Calc.xx`

This type of writing is mostly a personal preference I think.

Zaid-Ajaj: I think it makes it clear that I want the `Types` concerning the calc no other types. Keeping the same convesion for small and  large (nested) components. Other than that it has no significant value to choose one or the other so I will keep it this way :)

MangelMaxime: No problem :)

samples/react/ElmishCalc/src/State.fs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Active pattern that matches with an operation
+let (|Operation|_|) = function
```
forki: why not define a real DU?

forki: oh the DU is defined. then why do you need that AP?

Zaid-Ajaj: I am using it a lot inside the `update` function, so instead of checking the four operations every time, I check them all at once with it

forki: ah yes! makes sense

et1975:
Thanks!

PR #71, opened 2017-04-12
Added svg clock sample

Zaid-Ajaj:
This sample demonstrates the usage of svg in Elmish. However, I am not sure if I did the `Cmd` part idiomatically nor if there is a simpler way. Anyways here is it

elmish-svg-clock

### DIFF
--- a/samples/react/SvgClock/App.fs
+++ b/samples/react/SvgClock/App.fs
@@ -2,6 +2,7 @@ module App
 
 open System
 open Fable.Import.Browser
+open Fable.Core
 open Fable.Core.JsInterop
 
 // Types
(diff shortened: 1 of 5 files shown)


samples/react/SvgClock/App.fs
```diff
@@ -0,0 +1,85 @@
...
+open Elmish
+open Elmish.React
+
+
+let tickTime() = async {
+    do! Async.Sleep 1000
+    return DateTime.Now
+}
+
+let tickCommand = Cmd.ofAsync tickTime () Tick (fun _ -> Tick (DateTime.Now))
```
et1975: This should be a subscription, there is a sample in the native that does it.

et1975:
Neat, thank you!

Zaid-Ajaj:
@et1975 Thanks! 
Could you please explain to me why this didn't work:
```fsharp
let initialState() = CurrentTime DateTime.Now, Cmd.none
let update (Tick next) (CurrentTime time) = CurrentTime next, tickCommand
``` 
but when I changed `Cmd.none`to `tickCommand` in the `initialState()` it did work:
```fsharp
let initialState() = CurrentTime DateTime.Now, tickCommand
let update (Tick next) (CurrentTime time) = CurrentTime next, tickCommand
```

If it is only in the initial state, why didn't the command in the update take care of subsequent "ticks" unless I have passed first in the initial state?

et1975:
If there's no initial command, there's no call to `update`.

Zaid-Ajaj:
@et1975 that makes sense, thanks a lot :smile:

Zaid-Ajaj:
@et1975 Would you mind to take a look to see if it now looks good?

et1975:
Sweet, thanks!

Issue #76, opened 2017-04-13
Resolve depencies using paket


MangelMaxime:
@forki pointed me that thus dependencies are resolve by nuget.

```fs
<ItemGroup>
  <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />
  <PackageReference Include="FSharp.Core" Version="4.1.*" />
  <PackageReference Include="Fable.Core" Version="1.0.0-narumi-*" />
  <DotNetCliToolReference Include="dotnet-fable" Version="1.0.0-narumi-*" />
</ItemGroup>
```

We should use paket with a `.lock` file.

forki:
@alfonsogarciacaro @et1975 what do you guys think?

alfonsogarciacaro:
Hmmm, I think it's a bit overkill to use Paket for those 2 dependencies (I think the CLI tool and FSharp.NET.Sdk would still need to be in the project file) taking into account that Fable projects won't use any other Nuget dependency. I will undoubtedly use it for server-client projects like the fable-suave-scaffold (which we should make into a template soon).

But of course, if you think this is a good opportunity to promote Paket through Fable we can consider it :)

forki:
The thing is: it already failed to install on my machine because of
neverget

Am 13.04.2017 5:07 nachm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

MangelMaxime:
I myself add the same problem on the computer at work due to their nuget configuration.

So, it's could make sense.

alfonsogarciacaro:
I know, I know... my main concern is adding Paket to the template adds extra cognitive overhead for people who are not used to F# ecosystem. Paket brings many things with it:

- MONO dependency
- `paket.dependencies` and `paket.references` files
- `.paket folder` with extra files
- Auto generated: `paket.lock`, `packages` folder

And there would probably some questions pending:

- How do you run Paket? The convention is to use a build script, but this adds two more files (for Windows and Unix).
- Adding `Nuget.config` #75 didn't solve the issue?
- The Fable references (fable-elmish, fable-powerpack...) are also projects with the same Nuget references, to make Paket addition effective, we would have to add it as a convention for ALL Fable libraries.
- You would still have to use `dotnet restore` to get the FSharp.Net.Sdk and the CLI tool (dotnet-fable).
- The last point brings the other question, unfortunately adding Paket wouldn't remove Nuget completely which means that you could still have problems.

[...]

PR #79, opened 2017-04-13
Convert template to paket

forki:
references #76 
think about it...


forki:
Please do not merge yet. There is something missing

et1975:
This will need to go into templates project now.

PR #80, opened 2017-04-13
use an .proj to pack the template

enricosada:
it's a normal sdk proj (like the fsproj), so normal commands apply.
the extension matter for f#/c#, but any *.proj with `Sdk` attribute is a valid msbuild project for new sdk

good thing:

- need only .net core sdk
- normal proj sintax (no need to learn nuspec)
- normal commands and normal interoperability of msbuild (`/p`, env var, etc)
- Include and Exclude is easy

this also exclude by default the `node_modules` directory

```
cd src\templates\react\Fable.Template.Elmish.React.proj
dotnet restore 
dotnet pack
```

use `/p:Version=1.2.3`  to generate package as version 1.2.3
use `/p:PackageOutputPath=artifacts` to specify package output dir

or pass directy the path to .proj instead of change dir, as usual.

because is msbuild, you can choose how do things for properties:

- set env var `Version=1.2.3`
- add `<Version>1.2.3</Version>`
- pass `/p:Version=1.2.3`

same for `PackageOutputPath` if you want to set the relative path inside the .proj for example


enricosada:
@alfonsogarciacaro this replace the nuspec and fix the `node_modules` issue.

as a note, it's possible to create a package to remove the noise

```xml
      <PackageType>Template</PackageType>
      <PackProjectInputFile>$(MSBuildProjectFullPath)</PackProjectInputFile>
      <NoBuild>true</NoBuild>
      <IncludeBuildOutput>false</IncludeBuildOutput>
      <TargetFramework>netstandard1.0</TargetFramework>
      <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```

and replace it with

```xml
   <PackageReference Include="Fable.Template.Helper" PrivateAssets="All" />
```

to add defaults

enricosada:
@MangelMaxime i didnt found the template package generation in the fake script. but is a normal `dotnet restore + dotnet pack`

alfonsogarciacaro:
Bravo! You're setting the bar too high Enrico. Now we'll get impatient if you take more than two hours for a fix ;) I think the template publishing is not in the FAKE script because it involved downloading nuget.exe. I'll fix also the Fable simple template following this example.

MangelMaxime:
@enricosada Great thanks on this.
I see the exclude for `node_modules`. What would be the solution, if I need to exclude several folders, and even some generated files ? (exemple: obj, bin, public/bundle.js, etc.)

alfonsogarciacaro:
@MangelMaxime Maybe @enricosada can confirm but I think you can add several patterns to the `Exclude` attribute using semicolons:

```xml
<ItemGroup>
    <Content Include="Content/**/*"
                     Exclude="Content/node_modules/**/*; Content/public/bundle.js*; Content/bin/**/*" >
        <PackagePath>Content\</PackagePath>
    </Content>
</ItemGroup>
```

et1975:
Thanks!

enricosada:
yes, as alfonso said.

if is too long, you can also refactor extracting a property

```xml
<PropertyGroup>
      <ExcludeFromPackage>
          Content/node_modules/**/*;
          Content/public/bundle.js*;
          Content/bin/**/*
      </ExcludeFromPackage>
</PropertyGroup>
<ItemGroup>
    <Content Include="Content/**/*" Exclude="$(ExcludeFromPackage)" >
        <PackagePath>Content\</PackagePath>
    </Content>
</ItemGroup>
```

PR #82, opened 2017-04-14
Added Fifteen Puzzle sample

Zaid-Ajaj:
Here is one more :smile: [fifteen puzzle](https://en.wikipedia.org/wiki/15_puzzle)

fifteen-puzzle gif

### DIFF
--- a/samples/react/FifteenPuzzle/FifteenPuzzle.fsproj
+++ b/samples/react/FifteenPuzzle/FifteenPuzzle.fsproj
@@ -1,0 +1,30 @@
+﻿<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.6</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="node_modules\fable-powerpack\Fable.PowerPack.fsproj" />
+    <ProjectReference Include="node_modules\fable-elmish\Fable.Elmish.fsproj" />
+    <ProjectReference Include="node_modules\fable-elmish-browser\Fable.Elmish.Browser.fsproj" />
+    <ProjectReference Include="node_modules\fable-react\Fable.React.fsproj" />
+    <ProjectReference Include="node_modules\fable-elmish-react\Fable.Elmish.React.fsproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- Calc  -->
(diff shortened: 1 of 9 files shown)


et1975:
This is awesome, but to me the samples we host in elmish repo demonstrate an essential aspect of the lib, maybe you could host advanced samples like these in your own repo?
@alfonsogarciacaro  I think it's time we start awesome-fable list and link it from there.

forki:
I think we should create an official samples repo. I love the PR process
for samples, but yes maintenance of all samples doesn't need to be in main
repo. So we don't need to upgrade all the samples all the time.

Am 15.04.2017 3:17 vorm. schrieb "Eugene Tolmachev" <
notifications@github.com>:

alfonsogarciacaro:
Related to https://github.com/fable-compiler/Fable/issues/811

Yes, I was thinking about that for the issue above :point_up: Whether it was better to keep such a list in a page in Fable's web, a repo with a README or just a Wiki page (so community members can freely edit it).

That said, if we had a web page like the one @MangelMaxime did for fable-arch where samples can be easily added, I think apps like this one are a very good showcase that users can play with without installing anything. This is very important when programmers are comparing different frameworks and compile-to-JS languages in the internet. That web could be in this repo or a separate one for easier maintenance.

alfonsogarciacaro:
The other option, as fable-elmish is now our officially recommended library to build SPAs with Fable, is to have a section dedicated to fable-elmish in Fable website and put the sample there. I actually want to rewrite Fable website in elmish following fable-arch web model :+1:

Zaid-Ajaj:
So it comes down to two options: 
 - Create a repo with the samples inside the repo
   - Have the repos in there share dependencies vs. Having independent directories for each sample
 - Create a page (wiki, awesome-fable, docs site) and reference samples via links to their own respective repos
   - Every sample can then publish a demo for each repo in it's own gh-pages branch, this way you can reference a "live" preview from the docs site for example in an `iframe` just like @MangelMaxime was doing with fable-arch

To me, the second option is the way to go. What do you think?

forki:
Yes we need a separate repo that has a build script that allows to build all samples. And a deploy step that pushes this site to github pages. We then add back links to the original code on github and advertise the sample site like crazy. 

So people can learn about it and see code on github.

Ideally the sample site is subpage of the official docs.

et1975:
Now that we're in our own org, every member can create own a repo here. I'd like to see elm-like project structure, where a single sample/lib has a dedicated github project.

PR #83, opened 2017-04-14
For beginners: timer subscription sample

Zaid-Ajaj:
for beginners like me who needed to know how to subscribe to a timer to dispatch a message

### DIFF
--- a/samples/react/TimerSubscription/App.fs
+++ b/samples/react/TimerSubscription/App.fs
@@ -1,0 +1,48 @@
+module App
+
+open System
+open Fable.Import.Browser
+open Fable.Core.JsInterop
+
+// Types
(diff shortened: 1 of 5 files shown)


et1975:
Considering there's quite a bit of overlap with the SVG clock code, I think we are covered. A blog post and the sample hosted in your own repository would be sweet though!

Zaid-Ajaj:
Yes but my idea was that it is easier to have small-sized single purpose samples. If you are looking for a sample for time subscription, you would search for "time" and "subscription" rather than "svg" and "clock". 

Close this PR?

et1975:
Closing, let's follow one sample - one project structure like they do in elm-org.

PR #84, opened 2017-04-14
Fix #if DEBUG statement in the template

alfonsogarciacaro:
Reference:
https://github.com/dotnet/templating/issues/587
https://github.com/dotnet/templating/issues/432

This fixes the problem with the missing lines surrounded by `#if DEBUG` in the code generated by the template, unfortunately the `//-:cnd` comments will remain too, but apparently this will be fixed with dotnet SDK 2.0.

### DIFF
--- a/src/templates/react/Content/src/App.fs
+++ b/src/templates/react/Content/src/App.fs
@@ -74,7 +74,6 @@ Program.mkProgram init update root
 |> Program.withReact "elmish-app"
 //-:cnd
 #if DEBUG
-|> Program.withConsoleTrace
 |> Program.withDebugger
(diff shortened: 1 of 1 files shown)


et1975:
I don't think one would need both - debugger and console trace, especially if both are in conditional DEBUG block.

alfonsogarciacaro:
I thought that maybe it was convenient to have logs also in the browser console too, but that's ok. The important thing in this PR is to prevent the code within `#if DEBUG` disappears from the template :)

MangelMaxime:
Thanks @alfonsogarciacaro for the fix :)

PR #85, opened 2017-04-15
Updates all package.json to target "next" of Fable packages and samples.

mizzle-mo:
This might be too intrusive.  Please let me know.  I updated all package.json and yarn.lock files.

This will need to be changed to an exact version for RTM.

**NOTE:** This has not been tested.


src/elmish/package.json
```diff
@@ -16,6 +16,6 @@
   "author": "Eugene Tolmachev",
   "license": "Apache-2.0",
   "devDependencies": {
-    "fable-powerpack": "^1.0.0-narumi-3"
+    "fable-powerpack": "next"
```
MangelMaxime: I think we should keep this one fixed as this is the core library and not a sample.

et1975:
I'm going to close this for now, once the new project structure settles a bit we'll start accepting PRs again.

mizzle-mo:
Yeah my bad, it fell off the radar.  It seems like another change was made that supersedes this one as I am not having the issues with versions anymore.

PR #93, opened 2017-05-15
Publish Nuget package

alfonsogarciacaro:
This PR prepares the project to be published in Nuget instead of npm, to take advantage of the new Paket integration. Notes:

- The package will include both the .dll and [the sources](https://github.com/fable-elmish/elmish/blob/daad2136a71284b0dcec9d3f1729a7c199ef711e/src/Fable.Elmish.fsproj#L17-L21) in a `fable` package.
- `package.json` has been removed, as it's not necessary any more for publishing and this repo has not JS dependencies
- Paket is used (in magic mode) also to manage Fable.Elmish dependencies
- I haven't touched the build script, so you may want to include a FAKE target to publish the package. This are the necessary steps:

```
cd src
dotnet restore
# Set the version in the .fsproj file
dotnet pack
dotnet nuget push [PATH_TO_PACKAGE] -s nuget.org -k [YOUR_API_KEY]
```

> In this case it's not necessary to optimize the code in the .dll, so you don't need to do `dotnet pack -c Release`, because of this, the package will be placed in the `bin/Debug` folder by default.


et1975:
Do we care about embedding correct AssemblyVersion in this brave new fable world? cc @forki

et1975:
Also, should `paket` create the package instead of `dotnet`? cc @forki

src/Fable.Elmish.fsproj
```diff
@@ -18,6 +11,13 @@
     <PackageReference Include="FSharp.NET.Sdk" Version="1.0.*" PrivateAssets="All" />
     <PackageReference Include="FSharp.Core" Version="4.1.*" />
     <PackageReference Include="Fable.Core" Version="1.0.0-narumi-*" />
+    <PackageReference Include="Fable.PowerPack" Version="1.0.0-narumi-*" />
```
et1975: Does this get resolved using `paket`?

alfonsogarciacaro: Ups, yes. I forgot to delete that line 😅  Actually, FSharp.Core, Fable.Core and Fable.PowerPack references can be removed as they are now managed by Paket.

alfonsogarciacaro:
> Do we care about embedding correct AssemblyVersion in this brave new fable world? cc @forki

This is an interesting question. I find it very annoying that you cannot get the package version from the assembly (for dotnet-fable I'm using a literal automatically updated). It'd be nice but I guess it's not very important for Fable as you don't have access to assembly metadata in JS runtime.

src/Fable.Elmish.fsproj
et1975: I don't understand yet how `paket` interacts with the new `dotnet` tools, but if we just remove the `PackageReference`s, how would the compiler know to reference the Fable.PowerPack (for example)? Also, unless we switch to `paket` + `paket.template` and list the dependencies, how would we tell the consumer he needs that dependency?

et1975: Oh, looking at the `.targets` it seems to be overriding the targets for all of that. nm!

forki:
Yes paket generates PackageReferences in memory. So these should not be
manually added  in the fsproj

Am 16.05.2017 17:50 schrieb "Eugene Tolmachev" <notifications@github.com>:

et1975:
cc @forki 
`dotnet pack` fails for me (on OSX) with:
```
elmish/.paket/Paket.Restore.targets(66,5): error MSB4018: The "PackTask" task failed unexpectedly. [elmish/src/Fable.Elmish.fsproj]
elmish/.paket/Paket.Restore.targets(66,5): error MSB4018: System.IO.DirectoryNotFoundException: Could not find a part of the path 'elmish/src/obj/Fable.Elmish.0.9.0-beta-008.nuspec;elmish/src/obj'. [elmish/src/Fable.Elmish.fsproj]
elmish/.paket/Paket.Restore.targets(66,5): error MSB4018:    at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter) [elmish/src/Fable.Elmish.fsproj]
```

It looks like it concatenated some path that shouldn't be concatenated?

forki:
Yeah this looks like a bug in pack command. I'm on vacation until Sunday.
/cc @matthid @cloudroutine

Am 17.05.2017 12:25 vorm. schrieb "Eugene Tolmachev" <
notifications@github.com>:

matthid:
@et1975 can you create repro steps an issue on paket please?

PR #95, opened 2017-06-07
Add ofMsgs function to Cmd module.

generalelectrix:
Honestly a bit confused why this isn't here already.  May be an uncommon use case but I often find that I want to dispatch multiple messages from a single point and this is both more ergonomic and (probably) slightly more efficient (probably not importantly so).

Thanks for a nice, clean implementation of the Elm pattern.  Really enjoying it so far.

### DIFF
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -29,6 +29,10 @@ module Cmd =
     let ofMsg (msg:'msg) : Cmd<'msg> =
         [fun dispatch -> dispatch msg]
 
+    /// Command to issue a sequence of messages
+    let ofMsgs (msgs: #seq<'msg>) : Cmd<'msg> =
+        [for msg in msgs -> (fun dispatch -> dispatch msg)]
+
     /// When emitting the message, map to another type
     let map (f: 'a -> 'msg) (cmd: Cmd<'a>) : Cmd<'msg> =
         cmd |> List.map (fun g -> (fun dispatch -> f >> dispatch) >> g)


et1975:
Thanks, in spirit of Elm, I'm trying to keep it small. Having developed several fairly big apps with it, I can say it does feel like an antipattern, but to each his own. I assume you already know you can `batch`:
```fsharp
Cmd.batch [ Cmd.ofMsg Msg1; Cmd.ofMsg Msg2]
```

forki:
msgs |> Array.map Cmd.ofMsg |> Cmd.batch should do the trick

Am 07.06.2017 14:50 schrieb "Eugene Tolmachev" <notifications@github.com>:

Thanks, in spirit of Elm, I'm trying to keep it small. Having developed
several fairly big apps with it, I can say it does feel like an
antipattern, but to each his own. I assume you already know you can batch:

Cmd.batch [ Cmd.ofMsg Msg1; Cmd.ofMsg Msg2]

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/fable-elmish/elmish/pull/95#issuecomment-306784537>,
or mute
the thread

generalelectrix:
@et1975 Interested to hear more about your comments on Elm architecture as antipattern.

@forki Yep, I'm aware; the omission just felt a bit odd but really its due to the definition of ```Cmd<'T>``` as a list already, so there is no function available that I can pass to List.map and come out with ```Cmd<'T>```.  Just from regular F# habits the first thing I reached for was this function and it isn't present:
```
module Cmd =
    ...
    /// Wrap a 'msg as a Sub<'msg>.
    let sub = fun dispatch -> dispatch msg
```
Thus
```
[FooMessage; BarMessage]
|> List.map Cmd.sub
```
leaves one with a Cmd<Message>.

Not really important.

et1975:
Please don't misinterpret me, what I'm saying is spamming dispatch loop with messages lacks "mechanical sympathy":  single threaded JS runtime will handle all the updates as scheduled timeout (0) interrupts, which depending on the list size might get your app unresponsive. I've seen it happen, which is why I brought it up. On the other hand, restructuring the event to send a single message with a list of things might give *your own* logic a better chance at optimizing the handling of the list. Obviously there are times when either approach is more suitable than the other, that's why the `batch` is the first-class concept.

`Cmd.ofSub` is available, btw. Also note the fact that `Cmd` is a `list` is an internal implementation detail and might change in the future (like in Elm I might wrap that into an erased DU), so I wouldn't want it surfaced in the API.

generalelectrix:
Cool, gonna say this PR is definitely not necessary.

Small comment: as currently structured, the fact that Cmd is a list isn't properly hidden behind a newtype and functions does leak the implementation...

PR #97, opened 2017-06-07
Fix path for dotnet

MangelMaxime:
Current version, always download dotnet even if the right version is installed locally.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -16,7 +16,7 @@ let projects  =
 
 
 let dotnetcliVersion = "1.0.1"
-let mutable dotnetExePath = "dotnet"
+let mutable dotnetExePath = "./dotnetsdk/dotnet"
 
 let runDotnet workingDir args =
     printfn "CWD: %s" workingDir


et1975:
Hm, it doesn't do it for me on OSX... let me see, maybe we need to look it up like we do with `yarn`.

et1975:
I see what's happening, on my machine that's the version installed, so when it checks the version it finds the match, but if it's installed locally the script never tries to find it in the installed location.

CC @enricosada does anyone know what is significance of the dotnet version? Why do we need to check, cant we just try and run the build using whatever we find?

et1975:
@MangelMaxime I've switched to newer FAKE that has `DotNetCli` helpers to do this, can you check if it behaves correctly for you?

MangelMaxime:
It's ok for me. thanks

enricosada:
> CC @enricosada does anyone know what is significance of the dotnet version? Why do we need to check, cant we just try and run the build using whatever we find?
> View details

@et1975 it's because you usually want to use an exact version to build, so is deterministic.
If the right version is already installed, why download+unzip it (is lot of MB)?
Also the users can have any version installed (depends), like old project.json preview2 or new 2.0.

Maybe a better way is to say: use any `1.0.x` (no prerelease), but that's should be best done in fake helper.

alfonsogarciacaro:
I agree with @enricosada, it may be easier to allow SDK versions above the expected one so you don't have to download the zip just because you have a newer SDK installed. [This is what Fable does now](https://github.com/fable-compiler/Fable/blob/6c7519d23d1df38b1072a9ddff59b3bfeff57387/build.fsx#L295-L297), but it's true that we should also check `"./dotnetsdk/dotnet"` because sometimes the build script redownloads the SDK even if it's already there.

et1975:
The FAKE helpers do the smart thing and instead of installing into the working folder it installs into users' appData. Even though it's not added to the path it knows where to look for it next time it's needed.

PR #98, opened 2017-06-07
Subscriptions typo

blair55:
Make code sample reflect description


et1975:
Ha, good catch! 
However, it would fail to compile, `Program.mkSimple` does not actually expect tuple in the signature...
Let me see, perhaps I better update the description :)

et1975:
Updated the docs, what do you think?

blair55:
That's a clean, pure example.

One typo tho: convinience -> convenience

Thanks! :)

et1975:
Thank you, closing.

PR #99, opened 2017-06-12
Trace models after init and update. Also trace commands

forki:
* traces model after init
* traces msg before update
* traces model after update

Note: since model only changes through update or init we always have last model one line above in the log

### DIFF
--- a/src/program.fs
+++ b/src/program.fs
@@ -61,10 +61,24 @@ module Program =
     /// Trace all the updates to the console
     let withConsoleTrace (program: Program<'arg, 'model, 'msg, 'view>) =
         let inline toPlain o = toJson o |> JS.JSON.parse
-        let trace text msg model =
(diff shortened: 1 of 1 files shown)


et1975:
When I say "current state+msg is important" I mean it's important to have it within the same dispatch loop iteration. All kinds of things could end up in the log in between loop cycles and I wouldn't want the user to guess what the state was when this update call happened. I feel pretty strongly about this, but apparently I'm in minority as my colleague who does more of elmish than me agrees with you :)

forki:
Let's keep this in mind. I will report how this actually feels in practice
and I am happy to iterate on it. In the end we want to make it super useful
and I'm not married to a specific layout.

Am 12.06.2017 5:13 nachm. schrieb "Eugene Tolmachev" <
notifications@github.com>:

PR #110, opened 2017-09-12
[] -> Cmd.none

forki:
I think we should advertise Cmd.none instead of []


StachuDotNet:
Absolutely.
Relevantly, this little 'Cmd.none' will get a lot of hovertext usage. It may be worth adding some more information to https://github.com/fable-elmish/elmish/blob/master/src/cmd.fs#L24 for Elmish noobs to know what it *is* as many may consume Elmish without explicitly knowing how things work.

et1975:
Thanks!

Issue #119, opened 2017-10-18
The dispatch loop leaks memory when run on .NET


hvester:
### Description

The dispatch loop is not tail recursive (recursive call within a try with block) and thus leaks memory when run on .NET. See more details of a similar case here: https://stackoverflow.com/questions/24007897/f-mailboxprocessor-memory-leak-in-try-catch-block

### Repro code

Slightly modified basics.fsx to make the memory leak more apparent:

`#I "../../src/bin/Debug/netstandard1.6"
#r "Fable.Elmish.dll"

type Model =
    { x : int }

type Msg = 
    | Increment

open Elmish

let init () =
    { x = 0 }, Cmd.ofMsg Increment

let update msg model =
    match msg with
    | Increment -> 
        { model with x = model.x + 1 }, Cmd.ofMsg Increment

let view model _ =
    if model.x % 1000 = 0 then
        printfn "%i" model.x

Program.mkProgram init update view
|> Program.run`

### Expected and actual results

Steadily increasing memory usage when the above script is run.

### Related information 

* fable-elmish version: Latest master (900ef4010e9b863273f89fa7ae7c941015793617)
* .NET Core 2.0
* Operating system: Windows 10

et1975:
cc @2sComplement we'll need an update for the WPF version?

2sComplement:
Yea, was just looking at this. I'll bring it into Elmish.WPF, thanks!

PR #120, opened 2017-10-18
Fix memory leak in the dispatch loop

hvester:
Fix for the issue: "The dispatch loop leaks memory when run on .NET #119"


et1975:
Cool, thank you!

Issue #116, opened 2017-10-11
Cross-platform core


et1975:
There are now 2 non-fable builds of the same elmish core (that I know of) and I wonder if it's worth trying to accommodate both platforms and maybe make 2 different packages from the same codebase: netstandard and fable.

ctaggart:
@et1975 What are the other two non-fable builds that you know of. This is one:
https://github.com/fable-websockets/Fable.Websockets.Elmish/blob/master/src/Fable.Websockets.Elmish/src/Elmish.fs

I really enjoyed this video and I've watch it a couple of time.
https://youtu.be/1dJ-TlBr4Ro

Do you think it would be possible to combine Elmish with zeromq, perhaps with the pure .NET https://github.com/zeromq/netmq? I'm simply confused about the Elmish dispatcher loop and other loops and how they would interact. For example zeromq has a thread based [poller loop](http://learning-0mq-with-pyzmq.readthedocs.io/en/latest/pyzmq/multisocket/zmqpoller.html) and you can subscribe to events. How would we go about getting those events on to the dispatcher. In a code base I'm working on, we have an F# mailbox processor loop listening the netmq socket. Same question. I'm a bit lost in how these loops interact.

But yes, I think a netstand2.0 package would be great that doesn't require Fable.

et1975:
Yeah, sorry I haven't been able to dedicate any time to this.
@ctaggart [WPF](prolucid/Elmish.WPF) and [Xamarin](csoren/elmish-xamarin-forms). Some of my colleagues have done an internal project with Elmish.WPF and 0mq, not sure about the specifics, maybe @2sComplement can chime in.
I've been looking at various co-op async IO loops (dotnetty, tokio-rs, LMAX disruptor, etc)  and I can certainly see the appeal, but to me they have slightly different semantics/goals/priorities so I wouldn't try to integrate the two loops too deeply and just keep them running side-by-side and pass the messages in-between as needed. Elmish has a couple of options to `Sub` to external source of events, `Cmd.ofSub` and `Program.withSubscription` - the former can be issued from inside the `init` or `update` and your handler will get the up-to-date model, the latter will be called only once, with the initial model - otherwise the two are the same. How you'd publish the events from Elmish into those other loops will depend on the specifics obviously, but from Elmish perspective - you'd make the publish function available to the `update` as a command or as a function.

2sComplement:
[...]

PR #125, opened 2017-11-14
Remove console method calls

mavnn:
This pull requests removes calls to console.log and console.error from program.fs.

It does not fully remove the Fable dependency so that the `ofPromise` function can remain; a trade off which allows Elmish to be used in none Fable contexts, but recognizes that it is still primarily a Fable project.

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -30,7 +30,7 @@
     <DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
   </PropertyGroup>
 
-  <Target Name="PaketRestore" BeforeTargets="_GenerateDotnetCliToolReferenceSpecs;_GenerateProjectRestoreGraphPerFramework;_GenerateRestoreGraphWalkPerFramework;CollectPackageReferences" >
+  <Target Name="PaketRestore" Condition="'$(PaketRestoreDisabled)' != 'True'" BeforeTargets="_GenerateDotnetCliToolReferenceSpecs;_GenerateProjectRestoreGraphPerFramework;_GenerateRestoreGraphWalkPerFramework;CollectPackageReferences" >
 
     <!-- Step 1 Check if lockfile is properly restored -->
     <PropertyGroup>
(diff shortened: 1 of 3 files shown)


MecuSorin:
Another approach will be to shadow it with an implementation like:
```
module console =
    let log (label, msg) = System.Console.WriteLine("{0}{1}", label, msg)
```

et1975:
I considered using printf at some point early on, but the advantage of console.log is that it can show the arguments in a structured way. As @MecuSorin suggests, stubbing out certain fable namespaces is the way to go for now. See [Elmish.WPF](/prolucid/Elmish.WPF) for example. 
Setting up separate projects for Fable and .NET packages are in the backlog however - #116.

forki:
Why did you close it? This makes it work on .NET + fable

et1975:
I believe I gave a reasonable explanation, but here's another one: because if you read the contributor guidelines, I ask that you discuss the idea first. In the course of the discussion I could tell you that I made this very choice twice already and I that there's a way to addresses the cross-platform concerns in a way that doesn't sacrifice existing usability, and that's the PR I'd take gladly.

forki:
tbf @mavnn asked on twitter and I think @alfonsogarciacaro proposed to send this solution as PR ;-)

but yes in general there should probably be a new project for .NET eventually. But if I understand alfonso correctly then thisn would compile properly in both worlds

mavnn:
@forki  - @et1975 has a point. Asking the owner is a little different from following unsolicited twitter advice :)

The rational for keeping nice formatting of console logs makes sense, and taking a source reference and patching in the same way as Elmish.WPF would work in the short term.

alfonsogarciacaro:
Yeah, sorry, it was me who suggested to send the PR. That's the problem of mixing media for software discussions ;)

`Console.WriteLine(string)` will be compiled to `console.log` by Fable (if there are more arguments, Fable formatting will be applied). But I agree this is a Fable implementation detail and we shouldn't trust it forever (btw, `Console.Error` is not supported). Another solution for cross platform targeting is the `Emit` attribute:

```fsharp

type Console =
    [<Emit("console.log({0...})">]
    static member Log([<ParamArray>] args) = System.Console.WriteLine(args) // or similar
```

Fable will use `Emit` while .NET will use the body function. This won't remove the Fable.Core dependency although the attribute is recognized just by name so it'd be possible to define `Fable.Core.EmitAttribute` within Elmish, but this feels admittedly hacky.

The proper functional solution would be to pass the log function as a parameter but this would imply changing most of Elmish client. A less nice solution would be to make it a settable value somewhere (with `Console.WriteLine` as default) and let `Elmish.Browser` set it to `fun msg x -> console.log(msg, toPlain x)`.

PR #136, opened 2018-02-05
enable source link

ctaggart:


global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-      "version": "2.0.0"
+      "version": "2.1.4"
```
ctaggart: SourceLink requires > 2.0.2. 2.1.4 is the current SDK.

et1975:
Thanks!

ctaggart:
@et1975 It wasn't building on Travis CI yet.

paket.dependencies
```diff
@@ -3,6 +3,7 @@ source https://nuget.org/api/v2
 nuget FSharp.Core  redirects:force, content:none
 nuget Fable.Core
 nuget Fable.PowerPack
+nuget SourceLink.Create.CommandLine copy_local: true
```
et1975: Hm, should this be in the Build group instead?

netstandard/paket.references
```diff
@@ -1 +1,2 @@
-FSharp.Core
\ No newline at end of file
+FSharp.Core
+SourceLink.Create.CommandLine
```
et1975: Does it really need to be here?

src/paket.references
```diff
@@ -1,3 +1,4 @@
 FSharp.Core
 Fable.Core
-Fable.PowerPack
\ No newline at end of file
+Fable.PowerPack
+SourceLink.Create.CommandLine
```
et1975: Same here, I don't really want it to become a downstream package dependency.

paket.dependencies
ctaggart: Is there an issue with it there? The problem with the `Build` group was that it is limited to `framework: net46`.

netstandard/paket.references
ctaggart: Yes.

src/paket.references
ctaggart: It won't. That is what `copy_local: true` is for.

ctaggart: https://github.com/ctaggart/SourceLink/wiki/Paket

et1975: Right. Thinking further on this, Fable build will include sources anyways, so maybe we don't need to include the sourcelink here?

ctaggart: I don't understand what you mean by `Fable build will include sources anyway`.

ctaggart: Oh right. Fable is by source. Yes, you are correct.

ctaggart: But it doesn't hurt. If Visual Studio support F12 go to source at design time, it would be helpful.

et1975: I don't think I agree... I personally don't use VS, but I think I'd like it to go to local sources in a Fable project (as in "this is what you get when you compile this dependency"). Anyone else has a preference? cc @forki @alfonsogarciacaro @MangelMaxime

ctaggart: The source should be identical. Other editors should be able to add support besides VS.

MangelMaxime: I don't use VS so I don't know.

I would also prefer going to local version of the code if possible. As this would allow to not move from the editor.

I am already having this behavior with Ionide by moment but not sure if this is a feature or because, I use it to dev the libs.

alfonsogarciacaro: I think "Go to definition" works for Fable libraries developed by yourself because the file path is kept in the .dll. But it doesn't work for other libraries.

I think it'd be nice to be able to "Go to definition" for Fable libraries. SourceLink is magic for me as I don't know how it works :wink: but I assume it includes the sources in the package somehow. This may look as a "duplicate" as Fable libraries should already put the code in the `fable` directory, but it shouldn't add too much to the package size (which is cached by Nuget anyways) and I guess it'll be more work to make IDEs find the `fable` folder instead of using SourceLink, already an standard.

PR #137, opened 2018-02-05
Invert argument order in console trace

pirrmann:
I think it was forgotten in @vbfox 's work. The apps are not currently broken, but the console logs log the state instead of the message.

### DIFF
--- a/src/program.fs
+++ b/src/program.fs
@@ -67,9 +67,9 @@ module Program =
             Log.toConsole ("Initial state:", initModel)
             initModel,cmd
 
-        let traceUpdate msg model =
+        let traceUpdate model msg =
             Log.toConsole ("New message:", msg)
-            let newModel,cmd = program.update msg model
+            let newModel,cmd = program.update model msg
             Log.toConsole ("Updated state:", newModel)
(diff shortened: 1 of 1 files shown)


vbfox:
Ah yes 👍 I forgot that the update arguments were also inverted

src/program.fs
```diff
@@ -80,7 +80,7 @@ module Program =
     /// Trace all the messages as they update the model
     let withTrace trace (program: Program<'arg, 'model, 'msg, 'view>) =
         { program
-            with update = fun msg model -> trace msg model; program.update msg model}
+            with update = fun model msg -> trace msg model; program.update model msg}
```
pirrmann: I'm not sure it's relevant to invert the args here, it's a question of consistency (keeping the args always in the same order), versus backward compatibility.

vbfox: you mean invert them for the `trace` call yeah that's debatable. But might as well change them, breaking for breaking...

vbfox: (But it's @et1975  call 😄 )

et1975:
Thanks, that's on me.

PR #155, opened 2018-09-28
Fix Package target in build script

panesofglass:
When I attempted to run `build.cmd Package` locally, the build failed on the `Package` target. This PR resolves that problem for me.


et1975:
Thank you!

PR #156, opened 2018-10-15
Adding Cmd.exec

et1975:
I'm building a lib and I found `Cmd` as an API an interesting approach. 
This function provides a way to execute commands from a view, which was always possible in elmish, the function just makes it easier to accomplish.
@MangelMaxime thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ docs/temp
 Directory.Build.props
 temp
 paket-files
+.ionide
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ dotnet: 2.1.4
 dist: trusty
 
 before_script:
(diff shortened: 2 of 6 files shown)


Zaid-Ajaj:
I wouldn't recommend adding such functionality, it makes the update function dependent on the view, testability is already lacking when it comes to Elmish commands, this will makes things worse even when we have a nice way of simulating the update function in the unit tests

MangelMaxime:
I prefer to keep everything related to `Cmd` in the `update` part of the application.

I think people will miss use this functionality.

et1975:
Made it internal.

Zaid-Ajaj:
@et1975 Thanks a lot for understanding! Yes update is not the only source of effects but it usually is and it becomes more predictable if that is the case

Issue #157, opened 2018-10-27
Stabilizing dispatch instance in Fable


vbfox:
While trying to optimize react applications, passing dispatch in props of components like `PureComponent` or `React.memo` make them always update as the reference isn't currently stable (`mb.Post` results in a new call to `bind` every time a new message is received)

It can relatively easily be solved in an app by saving the dispatch instance and "stabilizing" it in the root view but that's code that every Fable use would need to use.

```fsharp
let mutable stableDispatch = None
let view model dispatch =
	let dispatch =
		match stableDispatch with
		| Some x -> x
		| None ->
			stableDispatch <- Some dispatch
			dispatch 
```

I think it should be solved in Elmish itself so it would just-work for everyone.

### Move all the code of `runWith` inside the mailbox

It has consequence on multi-thread applications as it change on what thread subscriptions are done so I don't know how compatible it is with non-fable elmish variants :

```fsharp
    let runWith (arg: 'arg) (program: Program<'arg, 'model, 'msg, 'view>) =
        let (model,cmd) = program.init arg
        MailboxProcessor.Start(fun (mb:MailboxProcessor<'msg>) ->
            let dispatch = mb.Post
            let rec loop (state:'model) =
                async {
                    let! msg = mb.Receive()
                    let newState =
                        try
                            let (model',cmd') = program.update msg state
                            program.setState model' dispatch
                            cmd' |> List.iter (fun sub -> sub dispatch)
                            model'
                        with ex ->
                            program.onError ("Unable to process a message:", ex)
                            state
                    return! loop newState
                }

            program.setState model dispatch
            let sub =
                try
                    program.subscribe model
                with ex ->
                    program.onError ("Unable to subscribe:", ex)
                    Cmd.none
            sub @ cmd |> List.iter (fun sub -> sub dispatch)

            loop model
        ) |> ignore
```

### Use a mutable variable to hold dispatch

The code is less "clean" on the elmish side but the change should be invisible to all consumers

```fsharp
    let runWith (arg: 'arg) (program: Program<'arg, 'model, 'msg, 'view>) =
        let (model,cmd) = program.init arg
        let mutable dispatch = ignore
        let inbox = MailboxProcessor.Start(fun (mb:MailboxProcessor<'msg>) ->
            let rec loop (state:'model) =
                async {
                    let! msg = mb.Receive()
                    let newState =
                        try
                            let (model',cmd') = program.update msg state
                            program.setState model' dispatch
[...]

Issue #130, opened 2018-01-26
Consider inverting model and dispatch arguments for the view


vbfox:
### Description

The current order of model and dispatch for the view function mean that the function can't capture dispatch and generate functions once.

I don't know for other Elmish implementations but for React it's a perf problem.

The code must generate action functions on each call of the view and it break all optimizations that something like PureComponent provide and force all DOM elements that have event handler to need to be changed, poluting the diff algorithm.

### Repro code

```fsharp
type [<Pojo>] BookProps = { key: string; id: string; book: string; removeBook: string-> unit }

type BookComponent(initialProps) =
    inherit PureStatelessComponent<BookProps>(initialProps)
	member this.removeBook() = this.props.removeBook this.props.id
    override this.render() =
        tr [] [
            td [] [ str this.props.book ]
            td [] [ R.a [ OnClick this.removeBook] [ str "Remove" ] ]
        ]

let view (model:Model) (dispatch: Msg -> unit) =
    let removeBook id = dispatch (RemoveBook id)
    table [] [
        model.WishList.Books
            |> List.map(fun book -> BookComponent { key = book.id; id = book.id; book = book.title; removeBook = removeBook })
            |> ofList
    ]
```

While we follow all of React best practices, the PureComponent is useless there as `removeBook` changes each time the view is called 😭 

### Expected and actual results

I expect the library to lend me in the pit of success and allow me to do get the best perf with ease 😉 

With the arguments inverted the view would look like that (With `Program.Run` changed to cache the result of passing just dispatch) :

```fsharp
let view (dispatch: Msg -> unit) =
    let removeBook id = dispatch (RemoveBook id)
    fun (model:Model) ->
        table [] [
            model.WishList.Books
                |> List.map(fun book -> BookComponent { key = book.id; id = book.id; book = book.title; removeBook = removeBook })
                |> ofList
        ]
```

React will always see the same props and optimize the hell out of it (And i'll ideally introduce a `BookListComponent` that is also pure so that `render()` is never called if books didn't change making it ultra cheap when books don't change.

### The dirty Hack (TM)

Some hack like that could work to make it resonably fast in the current state (No tested) but that's ugly:

```fsharp
type HackData<'msg, 'model, 'view> = {
    dispatch: 'msg -> unit
    captured: 'model -> 'view
}

let mkHackView (from: ('msg -> unit) -> 'model -> 'view): 'model -> ('msg -> unit) -> 'view =
[...]

PR #160, opened 2018-11-11
WIP: Ditching MailboxProcessor

et1975:
@MangelMaxime @vbfox @alfonsogarciacaro @Zaid-Ajaj 

Caveat: .NET/WPF will need to augment the  `dispatch` with synchronization on the UI thread, cause the new dispatch loop is not thread-safe.

Let's have some eyeballs on this - it seems to work interactively, but..
My elm-performance fork is not happy with it - the new TODO textbox appears to be filled out, but I get no  message about it, so it tries adding an empty TODO and `update` ends up ignoring it.


et1975:
Published as 3.0.0-beta-1

MangelMaxime:
TBH I don't really understand the code, so I can't help much for giving a review. 🤷‍♂️ 

Perhaps adding some comments to help explaining what is doing what can help if others want to look at the code. ?

forki:
what wwas the reason for getting rid of it?

MangelMaxime:
@forki This is related to
- https://github.com/elmish/elmish/issues/157
- https://github.com/elmish/elmish/issues/130

Small resume, it's for benefiting of React optimization out of the box. Currently we don't have them because we get a new referece on `dispatch` on each "loop" and react consider that it's a different function and so send an update for all our event handlers in the view.

forki:
So for .NET we could keep it and use compiler directive?

Am Mo., 12. Nov. 2018, 09:05 hat Maxime Mangel <notifications@github.com>
geschrieben:

et1975:
@forki the main issue is that currently MBP as implemented by fable does not guarantee FIFO for commands. As Maxime points out, there are benefits to React, but WPF would benefit from the dispatch loop running on the UI thread: https://github.com/elmish/Elmish.WPF/issues/40

`setState` can already be overriden to augment `dispatch` for the views, but I'll probably just introduce a `Program` property `syncDispatch` to do that, so that it works for commands as well. 

@MangelMaxime the ring is just a fancy queue (because there's no F# queue I'm having to implement one, so why not make it a fancy one ;)).

netstandard/Elmish.fsproj
```diff
@@ -8,7 +8,9 @@
   <ItemGroup>
     <Compile Include="../src/prelude.fs" />
     <Compile Include="../src/cmd.fs" />
+    <Compile Include="../src/ring.fs" />
     <Compile Include="../src/program.fs" />
+    <Compile Include="meta.fs" />
```
et1975: I can probably make do w/o this, we'll see if just referencing the ring source from the test project makes more sense that this jumping through the hoops.

src/program.fs
```diff
@@ -87,31 +87,35 @@ module Program =
...
-                        with ex ->
-                            program.onError ("Unable to process a message:", ex)
-                            state
-                    return! loop newState
-                }
-            loop model
-        )
-        program.setState model inbox.Post
+        let mutable rb = RingBuffer 10
+        let mutable reentered = false
```
et1975: flag indicating that dispatch is entered recursively via a cmd that was processed as part of earlier entry in the stack.

src/program.fs
```diff
@@ -87,31 +87,35 @@ module Program =
...
+                rb.Push msg
+            else
+                reentered <- true
+                let mutable nextMsg = Some msg
+                while Option.isSome nextMsg do // I wish we could write Rust-like `while (Some msg = nextMsg) do`
+                    let msg = nextMsg.Value
+                    try
+                        let (model',cmd') = program.update msg state
+                        program.setState model' dispatch
+                        cmd' |> Cmd.exec dispatch
```
et1975: This is where re-entry can happen and this is where Fable currently fails to FIFO.

tests/RingBuffer.fs
```diff
@@ -0,0 +1,44 @@
...
+let internal applyRBOps (ops:Op list) (rb:int RingBuffer) =
+    ops
+    |> List.map (function 
+                 | Pop -> 
+                   Poped <| rb.Pop()
+                 | Push v -> 
+                   rb.Push v
+                   Pushed)
+                   
+[<Property>]
```
et1975: This is FsCheck test, it generates 100 test cases by default, I think.

tests/RingBuffer.fs
```diff
@@ -0,0 +1,44 @@
...
+let actsLikeAQueue (ops:Op list) =  
+    let q = Queue()
+    let rb = RingBuffer 10
+    (q |> applyQOps ops) =! (rb |> applyRBOps ops)
```
et1975: Here we're asserting that the same operations on the queue and the buffer result in the same outcome.

et1975:
cc @cmeeren 

In .NET land you'll need to override `syncDispatch` on the program instance and set it to `Dispatcher.Invoke` or some such, because new dispatch loop is not thread safe.

drk-mtr:
@et1975 this is a bit of a long shot but I'm trying to work out why I'm seeing errors when running this: https://github.com/DieselMeister/Terminal.Gui.Elmish/tree/master/src/Terminal.Gui.Elmish/ElmishImplementation and I've just come across your comment.

I'm trying to determine whether trying your "override `syncDispatch`" suggestion might help, but I don't really understand how I'd do this. Would this involve adding another pipe when starting the program? e.g.

```fsharp
Program.mkProgram init update view
|> Program.withSyncDispatch xxx
|> Program.run
```

I wasn't able to find any reference to `Dispatcher` or `Invoke` within the Elmish repo so I've hit a bit of a dead end i.e. I've got no idea what the `xxx` would be (if I'm even on the right lines above that is).

If this involves actually implementing a dispatcher please ignore my query since that's probably beyond my capabilites at the moment!

et1975:
Hey @drk-mtr,

Since you have a copy of elmish in your code base you could just revert the
dispatch loop implementation to the one from the v2 branch:
https://github.com/elmish/elmish/blob/v2.x/src/program.fs#L88.
Alternatively, the syncDispatch implementation just needs to make sure only
one dispatch call happens at a time, something like:

let queuedDispatch dispatch =
    let inbox = MailboxProcessor.Start(fun (mb:MailboxProcessor<'msg>) ->
        let rec loop () =
            async {
                let! msg = mb.Receive()
                dispatch msg
                return! loop ()
            }
        loop ()
    )
    inbox.Post

bender2k14:
@drk-mtr, I implemented essentially the same thing as @et1975 in response to a similar question in #181.  (And now I cleaned up the code with @et1975's improvements, so it is exactly the same.)  See https://github.com/elmish/elmish/issues/181#issuecomment-530534973.

drk-mtr:
Huge thanks for the suggestions - really appreciate it!

PR #165, opened 2019-03-04
Update Fable.Core

alfonsogarciacaro:
The Fable.Core update is about the previous discussions for naming (Fable.Import.JS becomes Fable.Core.JS) and splitting Fable.Import.Browser into several Fable.Browser.XXX packages for easier maintenance.

I'm trying to send PRs for the other Elmish packages, please don't merge (or at least don't publish) until the other PRs are read :)


MangelMaxime:
If I understand correctly, this should need a major version of Elmish because of the new version of `Fable.Core` introducing breaking changes.

`Fable.Import.Js` --> `Fable.Core.Js` but perhaps I am messing with the SemVer I am not sure :)

alfonsogarciacaro:
Well, yes. But elmish 3 is still beta so it may be OK to update dependencies before the stable release.

et1975:
We'll need a readme update as well.

alfonsogarciacaro:
@et1975 I've updated the docs, what do you want to update in the README?

Also, I've removed the Fable.Promise dependency because it was not really necessary just to make two lines a bit more F# idiomatic :) But please let me know if you prefer to keep it.

I've sent similar PRs to elmish/browser and elmish/react. It'd be great if the PRs could be merged and new packages pushed so we can start testing them :+1:

et1975:
I can't recall right now what was wrong with the README, I'll review it and do the release of all this later.
Thanks!

PR #168, opened 2019-03-17
opaque Program, Cmd.Of*

et1975:
These are the ideas so far, I'll try applying some of this in Elmish.Browser and see if/how it feels.


src/cmd.fs
```diff
@@ -41,76 +37,203 @@ module Cmd =
...
-                task arg
-            with x ->
-                x |> (ofError >> dispatch)
-        [bind]
-
     /// Command to call the subscriber
     let ofSub (sub: Sub<'msg>) : Cmd<'msg> =
         [sub]
 
+    module OfFunc =
```
MangelMaxime: Is `Of` adding value compare to naming the module `Func`?

et1975: I think so - it is more discoverable by people who just start typing the same thing they are used to, and I'd want it in the name - function or module, the module seems preferable.

MangelMaxime: Ok, thanks for the explanations.

alfonsogarciacaro:
Is there any timeframe for releasing this? I'm planning to write a blog post announcing the changes in Fable.Core, Fable.Browser packages, etc. If Elmish API is going to change, it's probably a good idea to add information about that too, as it will likely affect most of Fable projects.

et1975:
I should be able to wrap it up in the next couple of days. Debugger turns out to be quite an outlier in the conversion.

et1975:
Released the changes along with Browser, React and Debugger. @MangelMaxime will you update the HMR?

MangelMaxime:
@et1975 Yes, I will do it this weekend.

alfonsogarciacaro:
No, you won't ;) https://github.com/elmish/hmr/pull/21

PR #169, opened 2019-03-19
Fix ofPromise signature

alfonsogarciacaro:
Sorry, when I changed the body of `ofPromise` I didn't notice the `onError` argument was inferred now to be `obj->unit`. I've fixed to prevent compilation errors when updating the library.


alfonsogarciacaro:
Would it be possible to release this? It seems I have rights, so I can do it if that's ok with you @et1975

et1975:
Go ahead, just keep in mind that I will merge #168 and then everything will be different anyway.

PR #193, opened 2019-11-01
Adds return type on all Promise functions

pauldorehill:
As per the title - the intellisense is a little confusing at the moment so just adds `Cmd<'msg>` as the return type to `perform` & `result`. There is one small issue I couldn't quite resolve: on the `Cmd.OfAsync`  functions the return type isn't aliased to `Cmd<'msg>` in VS or VSCode if running with Fable (in VS studio the initial pop up does show two signatures which is strange).


et1975:
Thanks!
Did you mean to target the master instead of v3 branch?

Issue #202, opened 2020-06-11
Documentation: Parent-child composition and user interaction


mjarosie:
Hi,

it's more of a topic about documentation improvement. I'd be willing to contribute, but my understanding of both Elmish and model-view-update architecture is still quite limited, so I thought I'd share my view so that the documentation could be improved which would (hopefully) benefit the project.

Bit of a background - I've just started learning functional programming and F#, I've never written a single line of Elm, but my first impression of the MVU approach is very positive! I'm describing my experience with the documentation from the point of view of someone who sees it for the first time. 

I went through the example of [parent-child composition](https://elmish.github.io/elmish/parent-child.html) and I've noticed that its content doesn't correspond to the UML sequence diagram on the [main page of the documentation](https://elmish.github.io/elmish/#Parent-child-composition-and-user-interaction) which is quite confusing. I think consolidating these two descriptions would make it way much clearer. I could modify either the diagram or the example, but because of my limited knowledge, I don't know which one of these two is more descriptive.

[I've implemented](https://github.com/mjarosie/elmish-parent-child-composition-demo/blob/master/src/App.fs) the logic described in [parent-child composition](https://elmish.github.io/elmish/parent-child.html) page and when going to http://localhost:8080 just the initial state of the application is being logged (after looking into it for a while I understand it's because messages are not being dispatched from the UI as in the example described on the main page of the documentation). I believe the interaction with the UI is not described because, as the documentation mentions, elmish is UI agnostic, but I'm wondering if maybe it should be framed more explicitly? For example, mentioning something along these lines at the end of [parent-child example](https://elmish.github.io/elmish/parent-child.html):

"To see the model working, implement the UI which dispatches `Counter.Msg.Increment` and `Counter.Msg.Decrement` messages so that the application can react to it."

(As I've mentioned, I don't have much experience with MVU so not sure if I described it correctly).

Alternative solution would be to refer to some sample implementation (something similar to other samples, such as [sample-react-counter](https://github.com/elmish/sample-react-counter)).

[...]

PR #203, opened 2020-06-14
Parent-child composition docs improvement

mjarosie:
Related to #202 

- fixed part of UML sequence diagram not being displayed
- improved parent-child composition description and made the description in [the main page](https://elmish.github.io/elmish/#Parent-child-composition-and-user-interaction) more consistent with the detailed walkthrough described in [parent-child example](https://elmish.github.io/elmish/parent-child.html)


et1975:
Thanks for the PR!
I might have some more comments once I pull this down on my machine and generate the docs, but can we keep it UI-neutral? I know most people land here because they are writing a web app, but I'd like to make it clear that Elmish doesn't depend on Html/React/etc, does not require a UI at all and that UI is not a primary focus of Elmish. 
React docs is where I think those specifics belong.

docs/content/parent-child-1-ui-init.wsd
```diff
@@ -0,0 +1,24 @@
+@startuml "parent-child-1-ui-init"
+' © Sebastian Porto 2016, 
+' licensed under a Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License.
+autonumber "<b>(#)"
+participant User #white
```
et1975: If you replace `participant` with `actor` it will show up as a stickman.

docs/content/parent-child-1-ui-init.wsd
```diff
@@ -0,0 +1,24 @@
...
+participant program #white
+participant Main.initialModel as Mim #white
+participant Main.view as MV #white
+participant Main.update as MU #white
+participant Widget.initialModel as Wim #white
+participant Widget.view as WV #white
+participant Widget.update as WU #white
```
et1975, line 12: Maybe no need to mention the update, since it's not invoked in this diagram?

et1975 (approved):
Actually it's fine, small changes suggested. Let me know if you want to update or I can merge as is.

docs/content/parent-child-1-ui-init.wsd
mjarosie: If I'll remove unused participants (`update` for both `Main` and `Widget` in the first image, `initialModel` in the second), images will have different widths (or one of these will be stretched more than the other). Here's how it looks when getting rid of unused participants:

[UI initialisation](http://www.plantuml.com/plantuml/uml/VP71IiD048RlUOeXFQZWbbR4WqXfwKN1Yb1ipfDaQ0TtPyDcqZwJh-6JcMs9BBHwpF_zyys4dRUAGJjd8MymaAYfTcnhCpKT6nRMFBk2xozOKucjCWgy-w0UxYRJXrlecUMAfAKQEgaf0C8o42hl2PRUEIyjB5G3bvso5_Fc9SOKAaPhrZiCjB3yIN0_di2BA0N1M4GBhoVnECDElNIkf03vKpcx7jta6LRg0toqVJOwx5WfQp0eLzoWA3J1RmEwIsY5BEFuNunsvMkoW2si-E_kdkbmx6mkLGgkjwHdmeBTF-r1MVmgi-5YCul7F48gpFeaWk8C514C8uUZF9demKdRGuSi9vHksa1A-i-u9OQfgm_CK7vM5tkfUqWJRtoM8Sxc97Ndx0y0)
[user interaction](http://www.plantuml.com/plantuml/uml/VP3DQi9058NtynGNNLJ1Y2sb2ocYkAdGb88ahcyo5nsSdp1pey_Kr-YJDJzDLB3TXNE-UywPB0EZvz9eyG3VNx2b30ChjF3fF3jucZs-JA3ojCh91f9GMaaU4DQUaDM9OEsCSJR0Yjchh6JbRFJXR2sJpnNgQ7j0JokjZWJFqnbiB9EtM8EeuRqDdWeisTdIPEHXy9ejHiFnG63EpaCIAcruFYWcKQ1dbQi2BKFXtTwZkMV5gEpqfEWC621E_qJAGY9J0oNte9sIU-8-QPV-0rsoTdsMQBP7Y_fo3ckjygEeFYiXJkUmiRad32JYMkhUCuTHcnY7VO-CnQ_Rt2Rt6J2eREmaQH4dJMi2bnOujnUDNpbHbzGEbzV-179d35evRfAkn_LuNzR3joEh5ryFIllVSBEZaZluZKtBNUewTKbMbaR_0000)

I think as these images are shown one above the other, that'll look better if we just leave "participants" lined up in both schemas.

docs/content/parent-child-1-ui-init.wsd
mjarosie: Done

mjarosie:
> 
> 
> but can we keep it UI-neutral? I know most people land here because they are writing a web app, but I'd like to make it clear that Elmish doesn't depend on Html/React/etc, does not require a UI at all and that UI is not a primary focus of Elmish.
> React docs is where I think those specifics belong.

Maybe that should be mentioned above where the user interaction is described? Something along these lines:

```
Following diagrams show interactions between components in case of a user interacting with a web app.

Note that Elmish doesn't depend on any specific UI such as HTML rendering (it actually does not require any UI at all), and HTML is used just for explanation purposes.
```

What do you think?

et1975:
Sounds good.

mjarosie:
Thanks for taking a look @et1975! Is there anything else that needs to be done from my side so that changes take effect?

Issue #162, opened 2019-01-05
To add the exit criteria or not


et1975:
cc @alfonsogarciacaro @MangelMaxime @Zaid-Ajaj @cmeeren 

I'm thinking to extend `Program` with an exit criteria. It would be just a predicate that tests for a specific value of '`msg',  i.e. `'msg -> bool' and be invoked by the dispatch loop to determine if it needs to exit. What do you guys think?

cmeeren:
I feel I'm lacking context and more details. In which situations is it helpful to control the exit of a loop? And have you considered other methods of exiting a loop?

et1975:
AFAIK currently there's no graceful way to stop the loop in JS. A typical SPA doesn't need to stop, but something that runs multiple instances of `Program`, or is running as an agent under node might benefit from this.
On .NET side terminating the process is possible of course, but that's a nuclear option when one might simply want to restart the loop at later time, maybe with different parameters.
Not that I have an immediate need for this, I just thought since we are already making changes as part of 3.0 I could put this in as well, but if the feeling is "no thanks" then that's ok too.

cmeeren:
I don't care much either way.

If implemented, extending `Program` with a `'msg -> bool` property seems like a good option. It means users can easily stop a loop from within the loop itself.

Another option, as @Nhowka suggests, is to provide a `unit -> unit` somewhere that users can call if needed, but I don't immediately see how this can easily be called for a specific `Program` instance/loop since the `Program` instance usually defined at the end of the compilation order and not available in the core Elmish logic.

Nhowka:
A way would be via a `ref`. Something like:

```fs
let killSwitch = ref ignore

Program.mkProgram init update view
|> Program.withKillSwitch killSwitch
|> Program.run
```

And then `!killSwitch ()` could be called.

cmeeren:
Yes, you'd need to rely on mutable state (though more or less official F# guidelines, such as "Expert F#", favor `mutable` instead of `ref`). Not particularly dangerous in this case, but I still like the message based approach a bit better.

MangelMaxime:
This remember me of an old issue I opened https://github.com/elmish/react/issues/3

And I guess if we had this feature at the time, then canceling a elmish program would have been possible by default without the need to re-implement the `withReact` function.

alfonsogarciacaro:
[...]

Issue #107, opened 2017-06-26
Program API doesn't force valid use of `with` modifiers


forki:
reraised from https://github.com/Prolucid/Elmish.WPF/issues/11

et1975:
Addressing this properly (static guidance, not runtime exceptions) would require completely different API, but for subscriptions I'm going to make the effect cumulative.

et1975:
Cumulative subscriptions released in 0.9.2

PR #227, opened 2021-11-07
Add exit criteria and handler

alfonsogarciacaro:
This adds the exit criteria and handler as implemented in `master` branch (v4 alpha) to v3. The only difference is, in order to [prevent a breaking change](https://github.com/elmish/elmish/issues/162#issuecomment-962466232), I've added a separate `mapTermination` helper instead of changing the signature of `map`. If this is accepted we could also deprecate `map` and add `mapInit`, `mapUpdate`...

An alternative implementation for termination could be to have a list of disposables (so subscriptions, extensions... can easily add a cleanup) and have `run` also return an IDisposable. That would make the API similar to those for observables and also avoids the need for program "managers" (like HMR or UseElmish) to have to wrap Msg in order to terminate the program.

cc @et1975 @MangelMaxime

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.301"
+    "version": "3.1.301",
+    "rollForward": "minor"
   }
 }
--- a/src/program.fs
+++ b/src/program.fs
@@ -16,6 +16,7 @@ type Program<'arg, 'model, 'msg, 'view> = private {
     view : 'model -> Dispatch<'msg> -> 'view
     setState : 'model -> Dispatch<'msg> -> unit
     onError : (string*exn) -> unit
+    termination : ('msg -> bool) * ('model -> unit)
     syncDispatch: Dispatch<'msg> -> Dispatch<'msg>
 }
 
@@ -34,6 +35,7 @@ module Program =
           setState = fun model -> view model >> ignore
           subscribe = fun _ -> Cmd.none
           onError = Log.onError
+          termination = (fun _ -> false), ignore
           syncDispatch = id }
 
     /// Simple program that produces only new state with `init` and `update`.
@@ -47,6 +49,7 @@ module Program =
           setState = fun model -> view model >> ignore
           subscribe = fun _ -> Cmd.none
           onError = Log.onError
+          termination = (fun _ -> false), ignore
           syncDispatch = id }
 
     /// Subscribe to external source of events.
@@ -88,6 +91,11 @@ module Program =
         { program
             with onError = onError }
 
+    /// Exit criteria and the handler, overrides existing. 
+    let withTermination (predicate: 'msg -> bool) (terminate: 'model -> unit) (program: Program<'arg, 'model, 'msg, 'view>) =
+        { program
+            with termination = predicate, terminate }            
+
     /// For library authors only: map existing error handler and return new `Program` 
     let mapErrorHandler map (program: Program<'arg, 'model, 'msg, 'view>) =
         { program
@@ -126,24 +134,36 @@ module Program =
           setState = mapSetState program.setState
           subscribe = mapSubscribe program.subscribe
           onError = program.onError
+          termination = program.termination
           syncDispatch = id }
+
+    /// For library authors only: map the exit criteria and handler
+    let mapTermination mapTermination (program: Program<'arg, 'model, 'msg, 'view>) =
+        { program with termination = mapTermination program.termination }
 
     /// Start the program loop.
     /// arg: argument to pass to the init() function.
     /// program: program created with 'mkSimple' or 'mkProgram'.
     let runWith (arg: 'arg) (program: Program<'arg, 'model, 'msg, 'view>) =
         let (model,cmd) = program.init arg
+        let toTerminate, terminate = program.termination
         let rb = RingBuffer 10
         let mutable reentered = false
         let mutable state = model        
+        let mutable terminated = false
         let rec dispatch msg = 
+          if terminated then ()
+          else
             if reentered then
                 rb.Push msg
             else
                 reentered <- true
                 let mutable nextMsg = Some msg
-                while Option.isSome nextMsg do
+                while not terminated && Option.isSome nextMsg do
                     let msg = nextMsg.Value
+                    if toTerminate msg then
+                        terminate state
+                        terminated <- true
                     try
                         let (model',cmd') = program.update msg state
                         program.setState model' syncDispatch


et1975:
Thanks for the PR Alfonso, but let's branch this off as v4 as this needs to be a breaking change. The `master` can be repurposed as v5 alpha.
The `map` is necessary because when changing generic arguments type we can't carry over any function that doesn't match the new generic constraint.

alfonsogarciacaro:
Oh, my, you're right. I totally missed that! Then, what do you think of adding termination through `IDisposable` to avoid issues with generics? There are a couple of drawbacks but maybe we can overcome them:

- The `IDisposable` returned from run won't go through the ring buffer so it needs to use its own lock to be thread-safe.
- The list of Disposables to be executed on termination won't have access to the model, but they can store their own state which should be enough in most situations (e.g. when creating a subscription with a time interval, you can also register a disposable with the id to clear the interval).

et1975:
I figure returning IDisposable from `run` would mean breaking change for
every elmish user, adding an additional mapper will only break library
authors.

alfonsogarciacaro:
Yes, sorry, what I was actually thinking of was to rename `run` to return the IDisposable and then provide `run/runWith` versions that just ignore the returned value to keep the same signature as it is now (like with `Observable.subscribe` and `Observable.add`). Example:

```fsharp
let startWith arg program: IDisposable =
    ...

let runWith arg program =
    startWith arg program |> ignore
```

> adding an additional mapper will only break library authors.

It's true that changing only `map` would not require no changes in the code for most users. But I'm concerned that changing `map` will automatically break any extension so all packages depending on Elmish would need to be updated (some users also have their own custom extensions), and sometimes the messages raised by these dependency errors [are obscure to the user](https://github.com/fable-compiler/fable-promise/issues/26). Also, if one library [needs termination capabilities](https://github.com/elmish/hmr/issues/29#issuecomment-962466927) they will have to update Elmish dependency but also keep maintaining the old version for users who haven't upgraded yet, which adds up to the maintenance load.

et1975:
Maxime already has working hmr and I have working navigation and debugging
that just need to be backported to be released as v4. This termination fits
nicely in update loop and lets not only extension but also the core app
logic decide if it needs to quit and how that should happen.
I'm not convinced that IDisposable would be lesser of two evils considering
between us three we probably have 80% of map userbase.
Maxime, what are your thoughts?

alfonsogarciacaro:
I've recently PRed/pushed several packages with a dependency on Elmish (Feliz.UseElmish, Fable.Lit.Elmish, Fable.Expect) and I'm also working on a big project with custom Elmish extensions. If we coordinate maybe we can update all the packages we own more or less at the same time and hopefully we do reach that 80% userbase :)

MangelMaxime:
Personally, I am in favour of making the changes as @et1975 proposed by changing the map function. 

If I remember correctly, the `map` function, has been created especially for cases like that making Elmish internals a "black box" with a strong contract exposed.

> Also, if one library needs termination capabilities they will have to update Elmish dependency but also keep maintaining the old version for users who haven't upgraded yet, which adds up to the maintenance load.

TBH we don't do a lot of maintenance on existing Elmish packages they are pretty stable in general.

Also, if people wants to use the latest feature they should just upgrade to Elmish v4. This is how we did in the past, and kind of the same way you did when upgrading Fable. You are not anymore actively working on Fable 1 or Fable 2. :) 

-----

When upgrading from Elmish 2 to Elmish 3, we prepared the release of all the packages in advance via beta release and when everything was working we released the stable version and it worked ok.

We just needs to take our time to test the beta versions making sure that everything works as expected. We can also try to makes an announce to inform libraries maintainer that they need to prepare for a new Elmish version.

------------

We also needs to decide what API changes we are going to introduce in the "new v4" are we just introducing the `termination` feature or are we also trying to introduce the rework of the `Cmd/Sub` systems from the master branch.

If we introduce only the `termination`, I believe only the package using program composition will be impacted. If we introduce `Cmd/Sub` systems rework it will impact more libraries because `Cmd` is now a real type and not just a list alias.

------------

Here is a first list of packages that needs to be updated:

- Fable.Elmish
- Fable.Elmish.React
- Fable.Elmish.HMR
- Fable.Elmish.Debugger
- Fable.Elmish.Browser
- Thoth.Elmish.Toast
- Feliz.UseElmish
- Fable.Lit.Elmish
- Fable.Expect
- Fable.Reaction ?

There are also some `Fable.Elmish.Remoting` ecosystem too but I never used it and so I don't know how it works.

et1975:
I've started backporting stuff to v3 to be released as v4 and I'm leaving
Cmd/Sub redesign in the master.

et1975:
Closing in favour of v4. @alfonsogarciacaro  4.0.0-beta-1 is released, please give it a spin?

alfonsogarciacaro:
Thanks @et1975! I gave it a quick try with Fable.Lit and seems to work: https://github.com/fable-compiler/Fable.Lit/commit/536b42eb3e10f3b6ac3fafc2878e7c15b08cbd75

It'd be nice to have a helper like this in the new version when you want to add a termination handler without overriding the exit criteria, and maybe also another to easily add disposable subscriptions:

```fsharp
let withTerminationHandler (handler: 'model -> unit) (program: Program<'arg, 'model, 'msg, 'view>): Program<'arg, 'model, 'msg, 'view> =
    program
    |> Program.map id id id id id (fun (criteria, handler2) ->
        criteria, fun model -> handler model; handler2 model)

let withDisposableSubscription (subscribe: 'model -> Dispatch<'msg> -> IDisposable) (program: Program<'arg, 'model, 'msg, 'view>): Program<'arg, 'model, 'msg, 'view> =
    let mutable disp: IDisposable option = None
    program
    |> Program.withSubscription (fun model ->
        Cmd.ofSub (fun dispatch ->
            disp <- subscribe model dispatch |> Some
        ))
    |> withTerminationHandler (fun _ ->
        disp |> Option.iter (fun d -> d.Dispose()))
```

Nhowka:
As the program terminates as soon as it receives the message and it doesn't have the chance to do a last update on the model, could we have the handler use that message as an argument? I believe it might simplify code where there is cleaning up after an unexpected error so we can send it to a logging service without needing to add it to the model and then sending an extra dummy message for the termination.

EDIT: Thinking more about that, what if instead of having it be a `'msg -> bool` and a `'model -> unit` we had a `'msg -> ('model -> unit) option`? This way we could have different handling logic for different termination messages and if the message content is needed, we can capture it on the returned function.

alfonsogarciacaro:
That's a good idea @Nhowka! I think it's good the terminate handler receives also the model. About mixing both the criteria and the handler, that would be convenient, but I guess it'll make it difficult to compose the terminate handlers. For example, if I wrap the Elmish program to terminate it at will (e.g. because of a hot reload or in React.useElmish because the component is removed from the DOM) I still want to call the terminate handler of the wrapped element.

```fsharp
let mapTermination (wrappedTermination: 'wrappedMsg -> ('wrappedModel -> unit) option) =
   function
      | Stop ->
           // I want to "activate" wrappedTermination but how do I do it?
           Some ignore
      | UserMsg msg ->
           wrappedTermination msg |> Option.map (fun f m -> Active m |> f)
```

Nhowka:
The solutions that I could think of are not that friendly. One of them would be `mapTermination` have an argument for the default `'wrappedMsg` termination message and the other would be a DU with one `('msg -> bool) * ('model -> unit)` for handlers that are always called on termination and terminate based on a criteria and one `'msg -> ('model -> unit) option` for handling termination that depends on the message. Maybe we could have different helpers for the simple criteria case and the derived one to hide that complexity from the end user and just expose it to library creators.

That could look like this:
```fsharp
let mapTermination (wrappedTermination: ElmishTermination<'wrappedMsg, 'wrappedModel>) =
   MsgBased <| function
   | Crash ex ->
      let handleExn ex = failwith "Implement logging of the exception"
      match wrappedTermination with
      | Conditional (_, handler) ->
          Some (fun m -> 
            handler m.WrappedModel
            handleExn ex)
      | MsgBased _ -> Some (fun _ -> handleExn ex)
    | UserMsg msg -> 
        match wrappedTermination with
        | Conditional (criteria, handler) ->
            if criteria msg then
                Some (fun m -> handler m.WrappedModel)
            else None
        | MsgBased handler ->
            handler msg |> Option.map (fun f m -> f m.WrappedModel)
```

et1975:
@alfonsogarciacaro I feel like I've made a mistake agreeing to make `withSubscription` accumulative, when the accumulation could and should have been left up to the user. Now we have some `with` functions that do and some that don't accumulate. I'd like to stop proliferation of this ambiguity and maybe actually reverse [this](https://github.com/elmish/elmish/issues/107#issuecomment-315427299).
I like the pattern the error handler uses: `with` overrides, `map` lets the user do accumulation.

I can see the appeal in `withDisposableSubscription`, but it would conflict with subscriptions redesign I'd like to have in vNext. Can we keep disposable subscriptions out of elmish core?

MangelMaxime:
I think it is possible to keep disposable subscription out of elmish core, because they can be created as an Elmish components.

For example, in one of my application, I have some event listener that are created at the components level and when the component is destroyed the listener are removed. I did it both using React components and from the update function so I don't think this is a hard limitation from Elmish right now.

alfonsogarciacaro:
> I like the pattern the error handler uses: with overrides, map lets the user do accumulation.

Yes, it makes sense to have the with- and map- prefixes have consistent behavior. It's not very obvious only from the name but I guess library authors will get it quickly. Removing accumulation from subscription may cause _silent_ breaking changes but we should be able to deal with it during the transition to v4.

And yes, it should be fine to leave the helper to add disposables subscriptions out of the core :+1: Given that `with-` will convey "override" meaning, and `map-` wrapping the current behaviour, should we agree on a prefix for accumulating when providing such helpers in libraries? `add-` looks like a good option:

```fsharp
let addTerminationHandler (handler: 'model -> unit) (program: Program<'arg, 'model, 'msg, 'view>): Program<'arg, 'model, 'msg, 'view> =
    program
    |> Program.mapTermination (fun (criteria, handler') ->
        criteria, fun model -> handler' model; handler model)

let addDisposableSubscription (subscribe: 'model -> Dispatch<'msg> -> IDisposable) (program: Program<'arg, 'model, 'msg, 'view>): Program<'arg, 'model, 'msg, 'view> =
    let mutable disp: IDisposable option = None
    program
    |> Program.mapSubscription (fun subscribe' model ->
        Cmd.batch [
            subscribe' model
            Cmd.ofSub (fun dispatch ->
                disp <- subscribe model dispatch |> Some
            )
        ])
    |> addTerminationHandler (fun _ ->
        disp |> Option.iter (fun d -> d.Dispose()))
```

et1975:
See #231 for the proposed changes.

Nhowka:
@alfonsogarciacaro Inspired by your use of mutable, I managed to implement a transformation that does what I imagined. But also, no need to be on core.

```fsharp
let withMessageTermination msgHandlerFactory program =
    let criteria, handler =
        let mutable handler = ignore

        let criteria msg =
            match msgHandlerFactory msg with
            | Some h ->
                handler <- h
                true
            | None -> false

        criteria, (fun m -> handler m)

    program
    |> Program.withTermination criteria handler
```

PR #230, opened 2021-11-21
Add Nacara for the documentation

MangelMaxime:
@et1975 

I ported the documentation from F# Formatting to Nacara. Sorry for the delay, I had to add [Literate F#](https://mangelmaxime.github.io/Nacara/nacara/guides/literate-files.html) support in Nacara in order for us to keep the docs in the `fsx` files.

## New requirements

The requirements to have Nacara running is to have at least Node.js 12.20, 14.14, or 16.0. 

The simpler, is to use the latest LTS version which is 16.13.0.

## Auto publish

It is possible to add a Github worflow to auto publish the documentation. I was not sure if that's what we wanted or no.

Example: 

https://github.com/fable-compiler/fable-promise/blob/5e41fb34ff59eda1b4e4bf8be389bd70038554c0/.github/workflows/publish.yml

## Not ported yet

I removed the "MODULES" section because I don't think people use it a lot. 

In the future, it will be covered by the API documentation generated by Nacara. I didn't add it right now, because it is still in beta and there are several syntax used by Elmish not well rendered yet.

Example 1:

image

Example 2:

image

The argument is missing `Program<...>`

Once, I have these issues fixed I will add it to Elmish.

## Possible improvements

Should we add a dropdown to the navbar redirecting to the different packages in the Elmish ecosystem?

This would helps the users navigates to the others packages.

Example from Nacara website:

image


et1975:
Thanks Maxime! I'm still testing the v4, hopefully we'll converge in the
next couple of weeks or so
I'm in favor of having links to the ecosystem projects and I don't have a
strong preference about auto publishing the docs, up to you.

docs/docs/basics.fsx
```diff
@@ -42,26 +53,36 @@ The `update` function will receive the change required by `Msg`, and the current
...
-    | Decrement when model.x > 0 -> 
-        { model with x = model.x - 1 }, Cmd.ofMsg Decrement
-    
-    | Decrement -> 
-        { model with x = model.x - 1 }, Cmd.ofMsg Increment
-
-(** 
-Again we return a tuple: new state, command. 
+    | Increment when model.Value < 3 ->
+        { model with
```
et1975, line 57: I'm skeptical that this is easier on the eyes than the single line expression.

docs/index.md
```diff
@@ -1,53 +1,61 @@
...
 * **Program** <br />
   This is an opaque data structure that combines all of the above plus a `setState` function to produce a view from the model.
   See the [`Program`](program.html) module for more details.
 
-
 ### Installation
 
-```shell
+```sh
 paket add nuget Fable.Elmish
```
et1975: We should probably replace this this with `dotnet package add` instead?

docs/docs/basics.fsx
MangelMaxime: TBH this is how I format my code now to have a consistant look every where between the declaration, update site, etc.

But, we can revert if you prefer. 

I tried to apply Fantamos to the files and the default of Fantomas don't seems really good for this file...

docs/index.md
MangelMaxime: Yes, we can use the `dotnet` command instead.

et1975 (approved):

MangelMaxime:
This week-end, I will port the other website to Nacara.

Like that we will have all the website looking the same and be ready for deployment with the new v4

et1975 (approved):

RELEASE_NOTES.md
```diff
@@ -1,165 +1,165 @@
-### 4.0.0-beta-2
```
et1975, line 8: I'm not sure if readme fake parser cares about ### vs ##, something to confirm.

MangelMaxime: I tried it, and it seems to be generating the `Meta.props` file correctly.

If it doesn't let me know, and we will just point to the Release notes, inside the source repository.

et1975 (approved):

et1975:
@MangelMaxime @alfonsogarciacaro I propose we cut the v4 release when the docs are ready, unless there's anything else holding us back.

MangelMaxime:
@et1975 I think we are ready, all the websites have been updated.

The API reference will comes later when Nacara support more F# types.

PR #231, opened 2021-12-01
Change withSubscriptions to override, add map functions

et1975:
Introduces breaking change to `withSubscription` - like all other `with` functions and F# record's `with` semantics it will replace current value instead of accumulating the subscriptions.
The user can accumulate with `Cmd.batch` or use new `mapSubscription` to implement aggregation of subscribers.


alfonsogarciacaro (approved):

et1975:
released in v4 beta-3

alfonsogarciacaro:
Thanks @et1975! It's working for Fable.Lit :+1: https://github.com/fable-compiler/Fable.Lit/pull/28

PR #232, opened 2021-12-14
Removed error with window.setInterval in subscription docs

tjausm:
The windows.setInterval function required a third argument.

### DIFF
--- a/docs/content/subscriptions.fsx
+++ b/docs/content/subscriptions.fsx
@@ -46,9 +46,9 @@ open Browser
 
 let timer initial =
     let sub dispatch =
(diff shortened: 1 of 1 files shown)


MangelMaxime:
A version with 2 arguments is available in `Fable.Core` module

Example from the future documentation: https://github.com/elmish/elmish/pull/230/files#diff-dc4a02b37d3c9e1268f051b2b357d32c34c782fcf75261b02f1a978df370c0feR126-R131

I think using this version is a bit easier for newcomers what do you think?

tjausm:
That is indeed a more elegant solution!

PR #241, opened 2022-05-23
[WIP] Don't merge: Compile with Dart

alfonsogarciacaro:
Some changes necessary to target Dart in the current state of Fable Snake Island compiler.

### DIFF
--- a/src/ring.fs
+++ b/src/ring.fs
@@ -3,15 +3,15 @@ open System
 
 [<Struct>]
 type internal RingState<'item> =
-    | Writable of wx:'item array * ix:int
-    | ReadWritable of rw:'item array * wix:int * rix:int
+    | Writable of wx:'item option array * ix:int
+    | ReadWritable of rw:'item option array * wix:int * rix:int
 
 type internal RingBuffer<'item>(size) =
-    let doubleSize ix (items: 'item array) =
+    let doubleSize ix (items: 'item option array) =
         seq { yield! items |> Seq.skip ix
               yield! items |> Seq.take ix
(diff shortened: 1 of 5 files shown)


src/ring.fs
```diff
@@ -3,15 +3,15 @@ open System
 
 [<Struct>]
 type internal RingState<'item> =
-    | Writable of wx:'item array * ix:int
-    | ReadWritable of rw:'item array * wix:int * rix:int
+    | Writable of wx:'item option array * ix:int
```
et1975, line 6: I understand the impulse to optionize, but the array is internal impl detail and there's nothing wrong with it using unchecked default for values because index range is tightly controlled anyway. From performance perspective having an option instead of value doubles the number of allocations and GC pressure.

alfonsogarciacaro: Yes this is a tricky thing with the Dart implementation right now. Dart has null safety so if I try to assign null to a generic it will complain and there's no such a thing as `default(T)` so I haven't found a way to allocate an array without providing a fill value.

At the moment, I'm compiling options as nullables in Dart so by using `'item option` I can assign null as default value. But I understand this has implications for .NET (would OptionValue help?) so I'm not asking to change it. Let's see if I manage to solve the issue in Dart, if not maybe we can also use `#if FABLE_COMPILER` here, although I'm also trying to avoid proliferation of conditional compilation :)

et1975: yes, `voption` should work and I think would be more or less free from perf/cost perspective.

et1975:
If someone wants to pickup this work we can reopen.

PR #248, opened 2022-08-25
subs receive current model, automatically started/stopped as needed

### LINKED ISSUE #183: Change subscriptions to receive current state of the model
I would like to propose to make the `SyncDispatch` more useful to library authors than it is at the moment 

## Background:
[Elmish.Streams](https://github.com/dbrattli/Elmish.Streams) replaces the default dispatcher with its own implementation (based on observables), which calls the default dispatcher internally. 

This makes it possible to send msgs to other places than the update function (in this case to a [stream function](https://github.com/dbrattli/Elmish.Streams/blob/c8341378cd4d72ba04172a5fbecac2382e3d1dc2/examples/Magic/src/Client/Magic.fs#L273).) This allows us to implement Elm like subscriptions that can be switched on and off based on the current model state and other very nice things.

Unfortunately currently only the dispatcher in the view can be replaced, so no msgs that are generated by commands can be fed into the stream function which makes it [a bit useless](https://github.com/dbrattli/Elmish.Streams/issues/25).

## Current state
Currently there are three ways to replace the dispatcher: `mapView` and `mapSetState` in `Program.map` and `Program.withSyncDispatcher`. 
It is not possible to use `mapSetState` because Elmish.React completely replaces `Program.setState` and is supposed to run at the end of the pipeline

I think  `Program.withSyncDispatcher` would be the way to go but this function call is not stable because `Program.map` always replaces the dispatcher with `id`, which is intended [to not fix the type of program when doing a map](https://github.com/elmish/elmish/issues/180)

This makes it very hard to use `withSyncDispatch` because you can never be sure if you have the dispatcher that you are expecting (it all depends if a library later in the pipeline uses map or not).

This situation is severed because Elmish.HMR uses `Program.map` in a way that is completely transparent to the developer (it is enough to open `Elmish.HMR`, [no need to put in in the pipeline](https://github.com/elmish/hmr/blob/8e5263451c8f23c069377e5c1de6d52235542268/src/hmr.fs#L152)).

This means currently it is not possible to have a custom dispatcher when `Elmish.HMR` is active and the code behaves differently when run in development and production.

### 1. Create a getter for `syncDispatch`
We could than do something like this:
```f#
program
|> Program.map blaba blabla
|> Program.withSyncDispatch (Program.syncDispatch program)
|> Program.runWith arg
```

Advantage: 
* No breaking change

Disadvantage: 
* every library that uses map needs to think of setting the dispatcher after doing a map, otherwise it might break other libraries
* the Program pipeline might even be more dependent on the order than it is now

### 2. add another parameter to `Program.map`: `mapSyncDispatch`
```fs
let map mapInit mapUpdate mapView mapSetState mapSubscribe mapSyncDispatch
        (program: Program<'arg, 'model, 'msg, 'view>) =
    { init = mapInit program.init
      update = mapUpdate program.update
      view = mapView program.view
      setState = mapSetState program.setState
      subscribe = mapSubscribe program.subscribe
      onError = program.onError
      syncDispatch = mapSyncDispatch program.syncDispatch }
```

Advantage:
* libraries are forced to provide a mapper 
* it is not possible to accidently discard a custom dispatcher

Disadvantage:
* breaking change

I would love to get some feedback about this.

kspeakman:
resolves #183 

This is a breaking change from v3.

After this change, subscribe works more like `view`. The user declares the subscriptions that should be running. Then Elmish automatically starts or stops subscriptions to make the running subs match the user's declaration.

Cool stuff
- Subscriptions can be changed automatically based on user interaction. See demo in linked issue.
- If so wired by the user, page-specific subscriptions are automatically started on page activation and stopped on deactivation.
- Easy to make new subscription providers. `Time.every` is ~5 lines of code.

What makes this possible is allowing users to choose IDs for subscriptions. Since subs run in the "background" across multiple updates, users will want to provide consistent IDs to keep the same subs running as long as the same conditions hold. This cannot be decided in an automated way by Elmish. Elm tried this. Example results: `Time.every` uses problematic shared timers (see elm/time 25). WebSockets platform support was [abandoned](https://github.com/elm-lang/websocket/blob/a4a62742089863abd2973fcd8afd0a580dca2068/README.md#history). Both in large part due to identity being a platform responsibility but ultimately undecidable.

## Changes

`subscribe` now returns a list of key/value tuples. The key is SubId (alias for `string`). The value is a new type defined like this:

```fsharp
type Sub<'msg> = { Subscribe: Dispatch<'msg> -> IDisposable }
```
_The previous type alias `Sub<'msg>` has been temporarily renamed to `Effect<'msg>` until permanent names are decided for both._

A `Subs.map` function is provided to map child msg subscriptions up to parents.

Subscriptions have been added to tracing.

### API design considerations

For the subscription, I used a record as a wrapper for the subscribe fn to make it easier to understand (vs a fn type alias). I considered creating another record type to contain the unsubscribe fn. However, IDisposable is a well-known .NET type used for this. But I have reservations about it. For subscription providers (e.g. `Time.every`) -- IDisposable feels a bit out of place compared to the rest of the API. Because it requires using a verbose object expression or creating a class with the IDisposable interface implemented.

I'm not sure if the ordering matters, but I am making sub changes after the user cmd is run. Just because it seems like the logical processing order from the user's perspective. (Principle of least surprise.) Subs are also stopped before calling user terminate. I'm assuming HMR will call terminate on hot reload, so maybe no extra handling for subs will be needed.

I'm not a big fan of making fns and data private in fp, so I used a module named Internal to signify it should not be relied upon as a stable API. Many times I wish library authors had given me that choice, and a few times I would have exercised it.

Anything is open for debate. I will ultimately defer to core maintainers/contributors.

### DIFF
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -12,11 +12,11 @@ open System
 /// Dispatch - feed new message into the processing loop
 type Dispatch<'msg> = 'msg -> unit
 
-/// Subscription - return immediately, but may schedule dispatch of a message at any time
-type Sub<'msg> = Dispatch<'msg> -> unit
+/// Effect - return immediately, but may schedule dispatch of a message at any time
+type Effect<'msg> = Dispatch<'msg> -> unit
 
-/// Cmd - container for subscriptions that may produce messages
-type Cmd<'msg> = Sub<'msg> list
+/// Cmd - container for effects that may produce messages
+type Cmd<'msg> = Effect<'msg> list
 
 /// Cmd module for creating and manipulating commands
 [<RequireQualifiedAccess>]
@@ -37,9 +37,9 @@ module Cmd =
     let batch (cmds: #seq<Cmd<'msg>>) : Cmd<'msg> =
         cmds |> List.concat
 
-    /// Command to call the subscriber
-    let ofSub (sub: Sub<'msg>) : Cmd<'msg> =
-        [sub]
+    /// Command to call the effect
+    let ofEffect (effect: Effect<'msg>) : Cmd<'msg> =
+        [effect]
 
     module OfFunc =
         /// Command to evaluate a simple function and map the result
--- a/src/program.fs
+++ b/src/program.fs
@@ -12,7 +12,7 @@ namespace Elmish
 type Program<'arg, 'model, 'msg, 'view> = private {
     init : 'arg -> 'model * Cmd<'msg>
     update : 'msg -> 'model -> 'model * Cmd<'msg>
-    subscribe : 'model -> (SubId * Sub<'msg>) list
+    subscribe : 'model -> Sub<'msg> list
     view : 'model -> Dispatch<'msg> -> 'view
     setState : 'model -> Dispatch<'msg> -> unit
     onError : (string*exn) -> unit
@@ -52,7 +52,7 @@ module Program =
     /// Subscribe to external source of events, overrides existing subscription.
     /// Return the subscriptions that should be active based on the current model.
     /// Subscriptions will be started or stopped automatically to match.
-    let withSubscription (subscribe : 'model -> (SubId * Sub<'msg>) list) (program: Program<'arg, 'model, 'msg, 'view>) =
+    let withSubscription (subscribe : 'model -> Sub<'msg> list) (program: Program<'arg, 'model, 'msg, 'view>) =
         { program with
             subscribe = subscribe }
 
@@ -76,7 +76,7 @@ module Program =
 
         let traceSubscribe model =
             let subs = program.subscribe model
-            Log.toConsole ("Updated subs:", subs |> List.map fst)
+            Log.toConsole ("Updated subs:", subs |> List.map (fun x -> x.SubId))
             subs
 
         { program with
@@ -88,7 +88,7 @@ module Program =
     let withTrace trace (program: Program<'arg, 'model, 'msg, 'view>) =
         let update msg model =
             let state,cmd = program.update msg model
-            let subIds = program.subscribe state |> List.map fst
+            let subIds = program.subscribe state |> List.map (fun x -> x.SubId)
             trace msg state subIds
             state,cmd
         { program with
--- a/src/program.fs
+++ b/src/program.fs
@@ -12,7 +12,7 @@ namespace Elmish
 type Program<'arg, 'model, 'msg, 'view> = private {
     init : 'arg -> 'model * Cmd<'msg>
     update : 'msg -> 'model -> 'model * Cmd<'msg>
-    subscribe : 'model -> Cmd<'msg>
+    subscribe : 'model -> (SubId * Sub<'msg>) list
     view : 'model -> Dispatch<'msg> -> 'view
     setState : 'model -> Dispatch<'msg> -> unit
     onError : (string*exn) -> unit
@@ -32,7 +32,7 @@ module Program =
           update = update
           view = view
           setState = fun model -> view model >> ignore
-          subscribe = fun _ -> Cmd.none
+          subscribe = fun _ -> []
           onError = Log.onError
           termination = (fun _ -> false), ignore }
 
@@ -45,13 +45,14 @@ module Program =
           update = fun msg -> update msg >> fun state -> state, Cmd.none
           view = view
           setState = fun model -> view model >> ignore
-          subscribe = fun _ -> Cmd.none
+          subscribe = fun _ -> []
           onError = Log.onError
           termination = (fun _ -> false), ignore }
 
     /// Subscribe to external source of events, overrides existing subscription.
-    /// The subscription is called once - with the initial model, but can dispatch new messages at any time.
-    let withSubscription (subscribe : 'model -> Cmd<'msg>) (program: Program<'arg, 'model, 'msg, 'view>) =
+    /// Return the subscriptions that should be active based on the current model.
+    /// Subscriptions will be started or stopped automatically to match.
+    let withSubscription (subscribe : 'model -> (SubId * Sub<'msg>) list) (program: Program<'arg, 'model, 'msg, 'view>) =
         { program with
             subscribe = subscribe }
 
@@ -73,15 +74,22 @@ module Program =
             Log.toConsole ("Updated state:", newModel)
             newModel,cmd
 
+        let traceSubscribe model =
+            let subs = program.subscribe model
+            Log.toConsole ("Updated subs:", subs |> List.map fst)
+            subs
+
         { program with
             init = traceInit 
-            update = traceUpdate }
+            update = traceUpdate
+            subscribe = traceSubscribe }
 
-    /// Trace all the messages as they update the model
+    /// Trace all the messages as they update the model and subscriptions
     let withTrace trace (program: Program<'arg, 'model, 'msg, 'view>) =
         let update msg model =
             let state,cmd = program.update msg model
-            trace msg state
+            let subIds = program.subscribe state |> List.map fst
+            trace msg state subIds
             state,cmd
         { program with
             update = update }
@@ -143,6 +151,8 @@ module Program =
           onError = program.onError
           termination = mapTermination program.termination }
 
+    module Subs = Subs.Internal
+
     /// Start the program loop.
     /// syncDispatch: specify how to serialize dispatch calls.
     /// arg: argument to pass to the init() function.
@@ -153,6 +163,7 @@ module Program =
         let rb = RingBuffer 10
         let mutable reentered = false
         let mutable state = model
+        let mutable subs = []
         let mutable terminated = false
         let rec dispatch msg = 
             if terminated then ()
@@ -165,26 +176,22 @@ module Program =
                     while not terminated && Option.isSome nextMsg do
                         let msg = nextMsg.Value
                         if toTerminate msg then
+                            Subs.stop program.onError subs
                             terminate state
                             terminated <- true
                         else                        
                             let (model',cmd') = program.update msg state
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
+                            subs <- program.subscribe model' |> Subs.recalc subs |> Subs.change program.onError dispatch'
                             state <- model'
                             nextMsg <- rb.Pop()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch            
 
         program.setState model dispatch'
-        let sub = 
-            try 
-                program.subscribe model 
-            with ex ->
-                program.onError ("Unable to subscribe:", ex)
-                Cmd.none
-        Cmd.batch [sub; cmd]
-        |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
+        cmd |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
+        subs <- program.subscribe model |> Subs.recalc subs |> Subs.change program.onError dispatch'
 
 
     /// Start the single-threaded dispatch loop.
--- a/src/program.fs
+++ b/src/program.fs
@@ -164,7 +164,7 @@ module Program =
         let rb = RingBuffer 10
         let mutable reentered = false
         let mutable state = model
-        let mutable activeSubs = []
+        let mutable activeSubs = Subs.empty
         let mutable terminated = false
         let rec dispatch msg = 
             if terminated then ()
@@ -177,7 +177,7 @@ module Program =
                     while not terminated && Option.isSome nextMsg do
                         let msg = nextMsg.Value
                         if toTerminate msg then
-                            Subs.stop program.onError activeSubs
+                            Subs.Fx.stop program.onError activeSubs
                             terminate state
                             terminated <- true
                         else                        
@@ -186,14 +186,14 @@ module Program =
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
                             state <- model'
-                            activeSubs <- Subs.recalc activeSubs subs' |> Subs.change program.onError dispatch'
+                            activeSubs <- Subs.getChanges activeSubs subs' |> Subs.Fx.change program.onError dispatch'
                             nextMsg <- rb.Pop()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch            
 
         program.setState model dispatch'
         cmd |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
-        activeSubs <- Subs.recalc activeSubs subs |> Subs.change program.onError dispatch'
+        activeSubs <- Subs.getChanges activeSubs subs |> Subs.Fx.change program.onError dispatch'
 
 
     /// Start the single-threaded dispatch loop.
--- a/src/program.fs
+++ b/src/program.fs
@@ -32,7 +32,7 @@ module Program =
           update = update
           view = view
           setState = fun model -> view model >> ignore
-          subscribe = fun _ -> []
+          subscribe = fun _ -> Sub.none
           onError = Log.onError
           termination = (fun _ -> false), ignore }
 
@@ -45,7 +45,7 @@ module Program =
           update = fun msg -> update msg >> fun state -> state, Cmd.none
           view = view
           setState = fun model -> view model >> ignore
-          subscribe = fun _ -> []
+          subscribe = fun _ -> Sub.none
           onError = Log.onError
           termination = (fun _ -> false), ignore }
 
--- a/src/program.fs
+++ b/src/program.fs
@@ -151,7 +151,7 @@ module Program =
           onError = program.onError
           termination = mapTermination program.termination }
 
-    module Subs = Subs.Internal
+    module Subs = Sub.Internal
 
     /// Start the program loop.
     /// syncDispatch: specify how to serialize dispatch calls.
@@ -186,14 +186,14 @@ module Program =
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
                             state <- model'
-                            activeSubs <- Subs.getChanges activeSubs subs' |> Subs.Fx.change program.onError dispatch'
+                            activeSubs <- Subs.diff activeSubs subs' |> Subs.Fx.change program.onError dispatch'
                             nextMsg <- rb.Pop()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch            
 
         program.setState model dispatch'
         cmd |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
-        activeSubs <- Subs.getChanges activeSubs subs |> Subs.Fx.change program.onError dispatch'
+        activeSubs <- Subs.diff activeSubs subs |> Subs.Fx.change program.onError dispatch'
 
 
     /// Start the single-threaded dispatch loop.
--- a/src/program.fs
+++ b/src/program.fs
@@ -12,7 +12,7 @@ namespace Elmish
 type Program<'arg, 'model, 'msg, 'view> = private {
     init : 'arg -> 'model * Cmd<'msg>
     update : 'msg -> 'model -> 'model * Cmd<'msg>
-    subscribe : 'model -> Sub<'msg> list
+    subscribe : 'model -> Sub<'msg>
     view : 'model -> Dispatch<'msg> -> 'view
     setState : 'model -> Dispatch<'msg> -> unit
     onError : (string*exn) -> unit
@@ -52,7 +52,7 @@ module Program =
     /// Subscribe to external source of events, overrides existing subscription.
     /// Return the subscriptions that should be active based on the current model.
     /// Subscriptions will be started or stopped automatically to match.
-    let withSubscription (subscribe : 'model -> Sub<'msg> list) (program: Program<'arg, 'model, 'msg, 'view>) =
+    let withSubscription (subscribe : 'model -> Sub<'msg>) (program: Program<'arg, 'model, 'msg, 'view>) =
         { program with
             subscribe = subscribe }
 
@@ -75,9 +75,9 @@ module Program =
             newModel,cmd
 
         let traceSubscribe model =
-            let subs = program.subscribe model
-            Log.toConsole ("Updated subs:", subs |> List.map (fun x -> x.SubId))
-            subs
+            let sub = program.subscribe model
+            Log.toConsole ("Updated subs:", sub |> List.map fst)
+            sub
 
         { program with
             init = traceInit 
@@ -88,7 +88,7 @@ module Program =
     let withTrace trace (program: Program<'arg, 'model, 'msg, 'view>) =
         let update msg model =
             let state,cmd = program.update msg model
-            let subIds = program.subscribe state |> List.map (fun x -> x.SubId)
+            let subIds = program.subscribe state |> List.map fst
             trace msg state subIds
             state,cmd
         { program with
@@ -159,7 +159,7 @@ module Program =
     /// program: program created with 'mkSimple' or 'mkProgram'.
     let runWithDispatch (syncDispatch: Dispatch<'msg> -> Dispatch<'msg>) (arg: 'arg) (program: Program<'arg, 'model, 'msg, 'view>) =
         let (model,cmd) = program.init arg
-        let subs = program.subscribe model
+        let sub = program.subscribe model
         let toTerminate, terminate = program.termination
         let rb = RingBuffer 10
         let mutable reentered = false
@@ -182,18 +182,18 @@ module Program =
                             terminated <- true
                         else                        
                             let (model',cmd') = program.update msg state
-                            let subs' = program.subscribe model'
+                            let sub' = program.subscribe model'
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
                             state <- model'
-                            activeSubs <- Subs.diff activeSubs subs' |> Subs.Fx.change program.onError dispatch'
+                            activeSubs <- Subs.diff activeSubs sub' |> Subs.Fx.change program.onError dispatch'
                             nextMsg <- rb.Pop()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch            
 
         program.setState model dispatch'
         cmd |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
-        activeSubs <- Subs.diff activeSubs subs |> Subs.Fx.change program.onError dispatch'
+        activeSubs <- Subs.diff activeSubs sub |> Subs.Fx.change program.onError dispatch'
 
 
     /// Start the single-threaded dispatch loop.
--- a/src/program.fs
+++ b/src/program.fs
@@ -159,11 +159,12 @@ module Program =
     /// program: program created with 'mkSimple' or 'mkProgram'.
     let runWithDispatch (syncDispatch: Dispatch<'msg> -> Dispatch<'msg>) (arg: 'arg) (program: Program<'arg, 'model, 'msg, 'view>) =
         let (model,cmd) = program.init arg
+        let subs = program.subscribe model
         let toTerminate, terminate = program.termination
         let rb = RingBuffer 10
         let mutable reentered = false
         let mutable state = model
-        let mutable subs = []
+        let mutable activeSubs = []
         let mutable terminated = false
         let rec dispatch msg = 
             if terminated then ()
@@ -176,22 +177,23 @@ module Program =
                     while not terminated && Option.isSome nextMsg do
                         let msg = nextMsg.Value
                         if toTerminate msg then
-                            Subs.stop program.onError subs
+                            Subs.stop program.onError activeSubs
                             terminate state
                             terminated <- true
                         else                        
                             let (model',cmd') = program.update msg state
+                            let subs' = program.subscribe model'
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
-                            subs <- program.subscribe model' |> Subs.recalc subs |> Subs.change program.onError dispatch'
                             state <- model'
+                            activeSubs <- Subs.recalc activeSubs subs' |> Subs.change program.onError dispatch'
                             nextMsg <- rb.Pop()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch            
 
         program.setState model dispatch'
         cmd |> Cmd.exec (fun ex -> program.onError (sprintf "Error intitializing:", ex)) dispatch'
-        subs <- program.subscribe model |> Subs.recalc subs |> Subs.change program.onError dispatch'
+        activeSubs <- Subs.recalc activeSubs subs |> Subs.change program.onError dispatch'
 
 
     /// Start the single-threaded dispatch loop.
--- a/src/sub.fs
+++ b/src/sub.fs
@@ -6,22 +6,30 @@ open System
 type SubId = string
 
 /// Subscription - Generates new messages when running
-type Sub<'msg> = { Subscribe: SubId -> Dispatch<'msg> -> IDisposable }
+type Sub<'msg> = { Start: SubId -> Dispatch<'msg> -> IDisposable }
 
 module Sub =
 
-    /// When emitting the message, map to another type
-    let map (f: 'a -> 'msg) (sub: Sub<'a>) : Sub<'msg> =
-        { Subscribe = fun subId dispatch ->
-            sub.Subscribe subId (f >> dispatch) }
+    module One =
 
-module Subs =
+        /// When emitting the message, map to another type
+        let map (f: 'a -> 'msg) (sub: Sub<'a>) : Sub<'msg> =
+            { Start = fun subId dispatch ->
+                sub.Start subId (f >> dispatch) }
+
+    /// None - no subscriptions, also known as `[]`
+    let none : Cmd<'msg> =
+        []
+
+    /// Aggregate multiple subscriptions
+    let batch (subs: #seq<Sub<'msg> list>) : Sub<'msg> list =
+        subs |> List.concat
 
     /// When emitting the message, map to another type.
     /// To avoid ID conflicts with other components, scope SubIds with a prefix.
     let map (idPrefix: string) (f: 'a -> 'msg) (subs: (SubId * Sub<'a>) list)
         : (SubId * Sub<'msg>) list =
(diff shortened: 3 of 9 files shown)


src/subs.fs
```diff
@@ -0,0 +1,54 @@
+namespace Elmish
+
+open System
+
+/// SubId - ID for live subscriptions, alias for string
+type SubId = string
```
et1975: We had a discussion around the implications of having globally-scoped ids under the direct control of a (sub)component in the application. 
As it is the guarantees we have so far in elmish that each child can be viewed and understood in isolation would no longer hold. Not only one would need to know what their component is doing, all other components and potentially libraries would be playing in the same Id space.
Maybe `map` should take the scope as additional parameter?
Maybe Id should be a member of `Sub`?

src/program.fs
```diff
@@ -143,6 +151,8 @@ module Program =
           onError = program.onError
           termination = mapTermination program.termination }
 
+    module Subs = Subs.Internal
```
et1975: I believe this is re-exporting the "internal" module as "public" `Subs` module, maybe it should just open?

src/program.fs
```diff
@@ -165,26 +176,22 @@ module Program =
...
                         let msg = nextMsg.Value
                         if toTerminate msg then
+                            Subs.stop program.onError subs
                             terminate state
                             terminated <- true
                         else                        
                             let (model',cmd') = program.update msg state
                             program.setState model' dispatch'
                             cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
+                            subs <- program.subscribe model' |> Subs.recalc subs |> Subs.change program.onError dispatch'
```
et1975: I'm curious about the placement, should it not follow `update` directly?

src/cmd.fs
```diff
@@ -12,11 +12,11 @@ open System
 /// Dispatch - feed new message into the processing loop
 type Dispatch<'msg> = 'msg -> unit
 
-/// Subscription - return immediately, but may schedule dispatch of a message at any time
-type Sub<'msg> = Dispatch<'msg> -> unit
+/// Effect - return immediately, but may schedule dispatch of a message at any time
+type Effect<'msg> = Dispatch<'msg> -> unit
```
et1975, line 16: The name is open of debate, from elmish perspective what's important is that it is a source of messages to be fed into the dispatch loop. Maybe `Feed`?
We also talked about cognitive overload of dealing with aliases. Yes, they make it easy to deal with functions, but they also make it harder to diagnose the errors compiler might give you.

et1975 (commented):
Thanks for the PR Kasey! Appreciate all the effort that went into this.
@MangelMaxime, @alfonsogarciacaro , @TysonMN, @jaggerjoe looking for your feedback.

src/subs.fs
kspeakman: ID scope parameter on `map` is a great idea. End users who have parent and child subs will quickly realize the need for this. For libraries it can be treated similar to model or msg: if the library needs its own subscriptions, it can prefix the child sub IDs to ensure no conflicts, otherwise pass it on.

I initially made `Sub` to include SubId. It's quite arguable, but I thought the record version was a bit verbose, and the tuple made it feel like the key/value that it is. Comparison below. Happy to change it.

```fsharp
// tuple version
let subscribe model =
    [ "subId1", Time.every model.Interval (Time.getNow Ticked) ]

// record version
let subscribe model =
    [ { SubId = "subId1" // line break for > 80 chars
        Subscribe = Time.every model.Interval (Time.getNow Ticked) } ]
```
_Could provide a helper fn to create the record more succinctly. But it will require extra parens: `[ Sub.mk "subId1" (Time.every ...) ]`, or a pipe syntax `[ Time.every ... |> Sub.mk "subId1" ]`._

src/program.fs
kspeakman: It's a module alias, stays local to the file. I can use an open if preferred. But didn't want to use `stop` and `change` unprefixed. And Internal was too long of a prefix for me.

src/program.fs
kspeakman: If you mean readability, yeah it would have been better to keep decisions (sub recalculation) above and effects below. I reckon I was trying to keep the footprint minimal. If you mean side effect order, my thought was principle of least surprise for the user. init/update is obviously run first, since the new model is needed for subscribe. So their effects are also run in that order.

src/cmd.fs
kspeakman: Looking at this abstraction from the source material Elm, each item in a Cmd list ends up as a `Task error msg`, which represents a single side effect and its return msg. The F# version of that would probably be `Async<'msg>`[^1].

An Elm Sub item turns into a Process, sortof an MVU app minus the view. It stays "running" until killed. Perhaps represented more economically in F# as `{ Start: unit -> IDisposable * IObservable<'msg> }`[^2].

Elmish's msg return mechanism is dispatch. Which expands the scope of what this abstraction (v3 Sub) does vs in Elm. The user can share out it's dispatch fn, by necessity making dispatch multi-use. So it's not quite a Cmd item because it doesn't represent a single effect and its return value[^3]. And it's not quite a Sub item because you have to explicitly manage its lifecycle side effects if you need other than app lifetime. It's made due for both concerns up to now. With v4 Subs in place, it wouldn't have to.

[^1]: To represent failure, the Msg case can be `of Result<>`, or different cases can be returned for failure/success. Elm uses Tasks for all side effects. But they have a special `Never` type to represent when a DU case is impossible. So `Task Never msg` is roughly equivalent to `Async<'msg>`.

[^2]: The mention of `IObservable<>` does not imply FRP program structure. It's used here because it's the best-fitting BCL abstraction for a Msg source (I can think of). `Dispatch<>` itself is a subset of the observer pattern in a function wrapper instead of an object one.

[^3]: In fact, it doesn't need to be in a list. Multiple v3 Subs can be combined by wrapping in a fan-out dispatch fn. `let batch subs = fun dispatch -> subs |> Seq.iter (fun sub -> sub dispatch)`

src/program.fs
kspeakman: See what you think now.

src/subs.fs
kspeakman: I added `idPrefix` as a parameter to `Subs.map`.

On Sub ID. I started to add it to Sub, but it seems to be at the wrong level of abstraction. Either the user will construct the record with SubId and be forced into extra structural detail (the Subscribe field). Or subscription provider unnecessarily pays the overhead of passing thru a sub id they can't use.

I considered cases where the subscriptions might want to use the sub id. Key-based features that need unique names, such as JS var names or storage keys. Putting SubId on Sub won't work for that. Sub creation only captures the original id. The subscribe fn can't access the updated version that may have been prefixed since then.

This leads me to think that SubId should be a parameter to Subscribe to support key-based features: `{ Subscribe: SubId -> Dispatch<'msg> -> IDisposable }` And also that SubId should still be kept separate from the subscribe fn. (Remembering that Sub is a record only to provide a Subscribe label.) Maybe SubId can be an SCU for easier user comprehension?

```fsharp
let subscribe model =
    [ SubId "a/b/c/d", Time.every 1000 (Time.getNow Ticked) ]
```

NB: SCUs are [terrible for perf](https://bartoszsypytkowski.com/writing-high-performance-f-code/#discriminatedunionsasvaluetypes) (bottom of section). Fable can erase them, but not dotnet AFAIK. Funny, but this would accomplish the same goal here: `let SubId (s: string) = s`.

et1975:
> NB: SCUs are terrible for perf
> https://bartoszsypytkowski.com/writing-high-performance-f-code/#discriminatedunionsasvaluetypes
> (bottom of section). Fable can erase them, but not dotnet AFAIK. Funny, but
> this would accomplish the same thing. let SubId (s: string) = s.

Marking SCU as [<Struct>] should eliminate the added GC pressure. I can't remember where I had it, but preprocessor-based conditional alias to either `Erase` or `Struct` attribute is how I've done this in the past.

src/subs.fs
```diff
@@ -0,0 +1,57 @@
...
+
+        let recalc (subs: (SubId * IDisposable) list) (newSubs: (SubId * Sub<'msg>) list) =
+            let keys = subs |> List.map fst |> Set.ofList
+            let newKeys = newSubs |> List.map fst |> Set.ofList
+            if keys = newKeys then
+                [], subs, []
+            else
+                let toKeep, toStop = subs |> List.partition (fun (k, _) -> Set.contains k newKeys)
+                let toStart = newSubs |> List.filter (fun (k, _) -> not (Set.contains k keys))
+                toStop, toKeep, toStart
```
kspeakman: Realized this doesn't screen for duplicates. This will lead to incorrect behavior when one of the dupes is turned off but the other isn't. Both will keep running. Preferably needs to error when there is a dupe. Implying "first one wins" behavior. Will fix.

kspeakman:
Struct SCU will add 4 bytes overhead for tag. (And another 4 bytes padding in this case.) Sleepy math. String: 8 bytes. Struct SCU of string: 16 bytes. Reference SCU of string: 24 bytes.

et1975:
Since structs are passed by value (and allocated on the stack) I'd say
that's not a problem.

kspeakman:
Agreed. Stack allocating and copying between stack frames does cost time/space vs copying an 8 byte reference. (It's why struct can lead to worse perf for larger and/or frequently passed types.) And I think struct SCU would lead to extra heap garbage because it's in gc tuple in a gc collection. But holistically, avoiding it for that reason here would be quite a micro-optimization. My spectrum traits showing, digging too deep into details. 🙃

I'm glad you mentioned this trick as I had not considered struct SCUs. I've been bitten before having two entities with similar names in the same use case and using the wrong ID. Non-obvious result behavior took a while to figure out. An SCU would have prevented that, and a struct SCU's lower overhead could make it worth doing.

src/subs.fs
kspeakman: Went with last one wins.

MangelMaxime:
I think having cancellable subscriptions would be a nice addition to Elmish.

About the implementation, I must admit that it feels a bit cryptic to me even with the explanation.

My options/ideas is that having the user provide the subscription is unfortunate but not that bad either. However, we should make sure the user can mess up the application easily. What I mean by that is that the user should not be responsible to ensure that the subscription is uniquer across the whole application loop. Mostly, because it probably don't know all the existing sub keys.

So if both the user and a library provide a subs called "tick" then there should be 2 subscriptions but user would just have called its `tick`
- `/root/libraryA/tick`
- `/root/.../userComponent/tick`

*I know that you discussed it above but I don't understand what it is the current state / agreement now.*

---

I find it strange that we are speaking about Subs/Subscriptions but there is something call `Effect` now. I find it confusing, either we need a consistent naming or be sure that the naming are correct and well explained to the user.

---

About HMR, if Elmish expose a way to call "dispose" on all the subscriptions then it should be possible to call that function on HMR call.

For example, in Elmish v3, Elmish.HMR contains a special code to remove the listener coming from Elmish.Browser [Source](https://github.com/elmish/hmr/blob/4d0fdca1d99e7f8401e4c3b038fb88917bebe895/src/hmr.fs#L200-L232)

kspeakman:
Thanks for the great feedback. 

> About the implementation, I must admit that it feels a bit cryptic to me even with the explanation.

Apologies for that. I'm open to suggestion. It might help to know that the impl uses a set-based diff algorithm to calculate what's changed from old to new subs. The general formula is this:

```fsharp
// assume old is set(a, b, c); new_ is set(b, c, d)
let removed = Set.difference old new_
// removed = set(a, b, c), take away b c d = set(a)
let added = Set.difference new_ old
// added = set(b, c, d), take away a b c = set(d)
let same = Set.intersect old new_
// same = set(a, b, c), elements in common with set(b, c, d) = set(b, c)
```

The impl uses List.partition to calculate "same" (toKeep) and "removed" (toStop). Then List.filter for "added" (toStart). It is logically equivalent to above, but avoids calculating the interim sets, then filtering lists by each of them.

I added a check for `keys = newKeys` to skip these calculations in a very common scenario.

In last commit, I added duplicate ID checking. This amounts to looping thru the new subs, adding their ids to a set, but first checking that they aren't already in the set. If they are, they go to the dupe list instead. The commit also adds tests for all the above described behavior.

> I know that you discussed it above but I don't understand what it is the current state / agreement now.

Regarding duplicate IDs. I added an `idPrefix` parameter to Subs.map. So this will be in the way / hard to miss when child subscriptions are mapped to use a parent Msg type. I added documentation explaining it. I feel the phrasing could be better, so I'm open to suggestions. Since last commit, duplicate IDs will error via program.onError.

> I find it strange that we are speaking about Subs/Subscriptions but there is something call Effect now. I find it confusing, either we need a consistent naming or be sure that the naming are correct and well explained to the user.

This naming is temporary until we iron that out. What would you prefer name-wise? The output of `subscribe` is a subscription list with subscribe/unsubscribe features. The output of `init`/`update` is model and ???, which provides a msg sending mechanism. Compared with Elm equivalents [here](https://github.com/elmish/elmish/pull/248#discussion_r956536378).

> About HMR, if Elmish expose a way to call "dispose" on all the subscriptions then it should be possible to call that function on HMR call.

> 
> For example, in Elmish v3, Elmish.HMR contains a special code to remove the listener coming from Elmish.Browser [Source](https://github.com/elmish/hmr/blob/4d0fdca1d99e7f8401e4c3b038fb88917bebe895/src/hmr.fs#L200-L232)

I spent a weekend studying Elmish.HMR, trying to figure out how to handle hot reloads. Made me very appreciative of your efforts. Looking back over `mapSubscribe`, it seems possible to make unsubscribe happen automatically on hot reload, once converted to this "live" subscription feature. I can help with that. Then any subscription will be auto stopped on hot reload w/o special handling. Navigation could even be one of these subscriptions instead of mutable/internal stuff. _I noticed the nav stuff was missing from master branch. Is that handling changed?_

kspeakman:
> About the implementation, I must admit that it feels a bit cryptic to me even with the explanation.

Wait, did you mean the public API? That's fair. I'm still trying to think of alternative ways to express it. Ideas welcome. The essential elements are an ID and a subscribe fn.

MangelMaxime:
>  I noticed the nav stuff was missing from master branch. Is that handling changed?

Master branch is supporting Elmish v4 which has termination support. So now, the navigation package is able to remove its listener itself without the need to do anything special in Elmish.HMR. At least, this is what is supposed now ^^

https://github.com/elmish/browser/blob/9913bdfd46d8867dc1dfdddbf72aeb71cb4f35a2/src/navigation.fs#L104-L115

---

About the cryptic implementation, this is not exactly related to your changes. Even if I understand Elmish a lot, I just don't understand most of the discussion that are occurring in the issues in general. 😂 I think this is a mix of not having the right vocabulary and not having studied CS which seems that sometimes I lack knowledge about some term, structures, etc.




> > I know that you discussed it above but I don't understand what it is the current state / agreement now.
> 
> Regarding duplicate IDs. I added an `idPrefix` parameter to Subs.map. So this will be in the way / hard to miss when child subscriptions are mapped to use a parent Msg type. I added documentation explaining it. I feel the phrasing could be better, so I'm open to suggestions. Since last commit, duplicate IDs will error via program.onError.

That's good to know and seems a good way to mitigate the problem.

kspeakman:
> So now, the navigation package is able to remove its listener itself without the need to do anything special in Elmish.HMR.

Awesome. Subs are stopped when the app terminates, so it should take care of itself.

> I just don't understand most of the discussion that are occurring in the issues in general.

I tend to get off in the weeds, like the SCU thing above. Sorry about that. For me, learning those concepts was more a matter of being old. 😆 Wonder what happened to that PHP app from ~2000.

kspeakman:
I think there's still some design room left on the subscription ID. Please look at [this gist](https://gist.github.com/kspeakman/21031ffde02869da95abee288a9ef21b).

MangelMaxime:
> Awesome. Subs are stopped when the app terminates, so it should take care of itself.

This was what I was going to write now. In theory, with Elmish v4 everything applications should cleans them self so for the HMR nothing special should be required to be done.

> I tend to get off in the weeds, like the SCU thing above. Sorry about that. For me, learning those concepts was more a matter of being old. 😆 Wonder what happened to that PHP app from ~2000.

Oh those application still work great I think.
Now days, I am wondering if we are not using SPA too much and should not go back to some good old multi pages application. I mean it was doing great for a lot of stuff in the past but that's a conversation to have somewhere else I think :)

TysonMN:
> ..., @TysonMN, ... looking for your feedback.

Oops.  I just noticed your request.

@kspeakman, do you have a demo using your code here where the UI is the command line?  If not, can you create such a demo in a GitHub repo and share a link to it?

I see some discussion above about examples of failed attempts to do "advanced" subscription features with automatically generated IDs.  Thus, the current proposal is for the user to provide the IDs.

I am not convinced that the correct solution is for the user to supply IDs.  I am wondering if the subscription function an return an object that makes all the "advanced" subscription features possible.  I will look at the CLI demo and see if I can make my idea work there.

kspeakman:
Hi @TysonMN. I probably need some help setting up such a demo. I spent a day trying to setup my other demo with the Elmish changes, but I believe I was using the wrong approach. Not quite sure where to start.

Did you see the [this](https://gist.github.com/kspeakman/21031ffde02869da95abee288a9ef21b) linked above? It goes into more details about what IDs do and no-user-ID solutions.

TysonMN:
Yes, I tried that demo.  It works great.  The ability to start, stop, and modify subscriptions is nice.

Now I want to understand how it works via a simpler demo.  Elmish is independent of the view.  It doesn't need to be a website like that demo, and it doesn't need to be a WPF application like with Elmish.WPF.  It suffices for the UI to be a CLI.  I expect such a demo will be easy to understand, review, and possibly change/improve.

kspeakman:
Ah, that link wasn't the demo. I'll include its contents below.

----

An early prototype of the "live subs" demo used the subscription settings + Msg type as the subscription ID. This worked as long as you only needed a single subscription. But it did not allow, for example, two timers on the same page with the same settings. That led me to switch to a user-provided Sub ID. This gives the user full control on when subscriptions are different or the same. It also gave ID the responsibility of triggering changes to a running subscription.

Subscriptions only have start/stop semantics. So "change" means stopping the running one and starting a new one with different settings. How ID controls this is illustrated in the [live subs demo](https://kspeakman.github.io/demo/livesubs/). Notice under Subscriptions -- which lists subscription IDs -- that the timer interval is part of the ID. When the user changes the interval from 1000 to 2000, the subscription ID goes from ".../interval/1000" to ".../interval/2000". This causes the old 1000ms subscription (JS interval) to stop and a new 2000ms sub to start. Conceptually there is only 1 app clock timer. But different settings makes it a different subscription instance.

When I switched to a fully user-provided Sub ID, it could be argued that too much responsibility was shifted to the Sub ID (and therefore the user). Like above, a change in subscription settings will usually mean the subscription is different. So the user will have to incorporate the same data into Sub IDs that they provide as subscription settings. It feels like duplicate work.

```mermaid
graph TD
    subgraph Timer Subscription
    SubId[Subscription ID]
    SubData[Subscription Settings]
    Model -- local ID --> SubId
    Model -- interval --> SubData
    SubData -- interval --> SubId
    end
```

I think @et1975 was trying to tell me this when we chatted. Per his suggestion, we could allow the user to choose the degree of responsibility. For example: an ID of `Auto` could have the subscription's ID depend only on the subscription settings. That would have worked for the demo clock. On the other end of the spectrum `Custom` could be fully user controlled (subscription settings ignored for ID purposes). There's room in between for the combination of a local ID + subscription settings, so the user can differentiate two subscriptions locally, like "timer1" and "timer2", but they will automatically restart when the settings change. Probably the easiest way to do this would be to have the sub provider stringify the subscription settings (maybe provide some helpers), which would get combined with the local ID. This all assumes assumes that IDs are scoped/prefixed to their owner, which idPrefix is doing for us when mapping.

Thoughts?

Nhowka:
> When I switched to a fully user-provided Sub ID, it could be argued that too much responsibility was shifted to the Sub ID (and therefore the user)

I believe that is a good thing because, most of the time, that is something that can be abstracted away. Like in the interval example, that could be used in a `Cmd.ofTimer interval action` extension. I don't believe that ID is burdening the user, but it's giving freedom to the library authors.

src/program.fs
et1975: Right, I don't have a strong preference atm, just trying to figure the intent. To me `update` and `subscribe` are about generating the inputs and `setState` and `exec` are about affecting the change, which is why I thought I'd place them together.

kspeakman:
> I don't believe that ID is burdening the user, but it's giving freedom to the library authors.

I am of the same mind. Even as an end user, sometimes I have advanced scenarios. I want the freedom to handle those without going into technical debt. Aka, working around arbitrarily imposed limitations.

Libraries can add APIs to make it nice for the user. I guess the pertinent question is how much responsibility do we want to put into the Elmish library. So far, it just expects keys, and takes care of the diffing to start and stop subs.

et1975:
Adding the scope to the map function solves all my concerns about IDs. I agree that there's no foolproof way to only autogen them and would not even bother with auto-stringification based on the inputs.

kspeakman:
@TysonMN Would it be sufficient to create a CLI port of the demo? There are some things I'm not sure how to handle, but I could package all the necessary pieces together into a repo.

src/sub.fs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Subscription - Generates new messages when running
+type Sub<'msg> = { Start: SubId -> Dispatch<'msg> -> IDisposable }
+
+module Sub =
+
+    module One =
+
+        /// When emitting the message, map to another type
+        let map (f: 'a -> 'msg) (sub: Sub<'a>) : Sub<'msg> =
```
et1975: This is why I was wondering if SubId should be in Sub - here we don't have the way to add prefix.

src/sub.fs
```diff
@@ -0,0 +1,88 @@
...
+            { Start = fun subId dispatch ->
+                sub.Start subId (f >> dispatch) }
+
+    /// None - no subscriptions, also known as `[]`
+    let none : Cmd<'msg> =
+        []
+
+    /// Aggregate multiple subscriptions
+    let batch (subs: #seq<Sub<'msg> list>) : Sub<'msg> list =
```
et1975: Should have same return type between `none` and `batch` imo.

kspeakman: Made some copy pasta there. 🙃

src/sub.fs
kspeakman: I think you're right about the ID. Especially with what I'm about to post.

kspeakman:
This is what `subscribe` usage could look like to the user (if these subscription libraries existed).

```fsharp
let subscribe model =
    //if model.Terminated then Sub.none else
    Sub.batch [
        // app-level examples
        [ Navigation.hashBased HashChanged
          WebSocket.connection model.NotifySettings Notified
          match model.Timer with
          | On intervalMs ->
              Time.every intervalMs (Time.now Tick)
          | _ -> () ]
        // ActivePage matches page case, calls page subscribe, maps Msg
        ActivePage.subscribe model.ActivePage
    ]
```

This makes the subscription provider responsible for the ID. But per the previous post, usually some subscription settings control whether the sub should restart. So these could be provided as a default ID. For example:

```fsharp
module Time =

    let every (intervalMs: int) (effect: Effect<'msg>) : Sub<'msg> =
        "interval/" + string intervalMs,
        { Start = fun subId dispatch ->
            let handler () = effect dispatch
            let intervalId = window.setInterval(handler, intervalMs)
            { new IDisposable with
                member x.Dispose () = window.clearInterval(intervalId) } }
```

This would cover a common case where the user only needs one timer on a given page. And it provides auto-restart functionality if the interval is changed. To handle multiple timers within the same scope, Time could provide another every-type fn accepting an ID. Or Sub could provide a pipeable `withId` fn.

```fsharp
let subscribe model =
    [ Time.every 1000 (Time.now Tick) |> Sub.withId "tick"
      if model.IsPolling then
          Time.every 1000 (Time.now Poll) |> Sub.withId "poll" ]
. . .

module Sub =
    let withId idPrefix (subId, sub) =
        sprintf "%s/%s" idPrefix subId, sub
```

Sub IDs would be `["tick/interval/1000"; "poll/interval/1000"]` before being mapped to parent.

Edit, forgot to summarize

This gives the library the ability to abstract away the ID. Or to ask the user for some/all of the ID if it makes more sense for the library's use cases. When the user faces advanced scenarios, they can override or augment the ID provided by the lib.

Thoughts?

src/sub.fs
kspeakman: I was trying to get rid of the Sub/Subs distinction like Cmd, but it feels like I'll have to bring it back.

kspeakman: The one-letter difference between Sub and Subs is too subtle I think. Especially if the same named fn `map` is going to exist for both. I could just remove the individual map. Cmd takes this approach, there was no map fn for an individual item.

kspeakman: Or do you like the `Sub.One.map` API?

et1975: I'd be ok removing `One` module entirely and merging `Subs` into `Sub`. I think for Cmd it didn't even come up because plural was impl detail hidden behind the alias.

TysonMN:
> @TysonMN Would it be sufficient to create a CLI port of the demo? There are some things I'm not sure how to handle, but I could package all the necessary pieces together into a repo.

Yes, I think so.

src/sub.fs
kspeakman: Fixed now.

tests/SubTests.fs
```diff
@@ -0,0 +1,95 @@
...
+    let toKeys keyValueList =
+        List.map fst keyValueList
+    let toKeys2 subs =
+        subs |> List.map (fun sub -> sub.SubId)
+    let toIds (dupes, toStop, toKeep, toStart) =
+        dupes, toKeys toStop, toKeys toKeep, toKeys toStart
+    let toIds2 (dupes, toStop, toKeep, toStart) =
+        toKeys2 dupes, toKeys toStop, toKeys toKeep, toKeys2 toStart
+    let run = Sub.Internal.diff
+    let eq expected actual (message: string) =
```
et1975: Unquote lib provides nice `=!` operator that reports expected and actual, if equality assertion fails.

kspeakman: Ah. I'll give it a try. I know very little of test libs. Just enough for tests like these.

kspeakman: I ended up keeping eq to avoid repeating `toId` fns in every test. But it now uses `=!` so the data can be seen on failure.

kspeakman:
@TysonMN Here you go. [repo](https://github.com/kspeakman/ElmishSubCliDemo). You'll have to fill in the console interaction part. There's a readline in Main.fs keeping it running.

TysonMN:
Thanks. I will take a look within the next few days.

src/program.fs
kspeakman: I believe I resolved this. If not please tell me. Side note: I can see why v3 runs subscribe effects first, and my clock demo should have. Otherwise, user msgs queue first, update runs before sub start/stop confirmations are processed, more sub start/stop effects triggered. Causing dupe running subs.

src/subs.fs
kspeakman: Let me know if this needs further work. @et1975

kspeakman:
I've noticed in using this that I feel the uncomfortable need to pick a separator character between ID segments. Perhaps a better approach would be to make the ID a list of strings. Prefixing is then just `idPrefix :: subId`. This would also make it easier to work with if a library or user needs to adjust the ID.

src/sub.fs
```diff
@@ -0,0 +1,90 @@
...
+
+open System
+
+/// SubId - Subscription ID, alias for string list
+type SubId = string list
+
+/// Subscription - Generates new messages when running
+type Sub<'msg> =
+    { SubId: SubId
+      Start: SubId -> Dispatch<'msg> -> IDisposable }
```
et1975: Why do we need to pass SubId into Start?

kspeakman: Seemed like a good idea at the time, but now I’m struggling to come up with an example where I would use it. I’ll take it out when I get back to a computer, Mon.

TysonMN:
> @TysonMN Here you go. [repo](https://github.com/kspeakman/ElmishSubCliDemo). You'll have to fill in the console interaction part. There's a readline in Main.fs keeping it running.

Oh, I just noticed that you asked me to do the interaction part.  I would prefer that you do this.  Otherwise, I might not your your new API as you intend.

kspeakman:
> Oh, I just noticed that you asked me to do the interaction part. I would prefer that you do this. Otherwise, I might not your your new API as you intend.

I'm not familiar with using Elmish from the console. From your request, I thought you would be. To add this interaction, I would basically be creating an Elmish.Console platform integration, which seems non-trivial. I don't understand how existing integrations would be validated by demonstrating a new and completely different integration.

If we're ultimately talking about Elmish.Wpf, the [Elmish.LiveSubs](https://github.com/kspeakman/ElmishSubCliDemo/tree/main/src/Elmish.LiveSubs) project should be directly usable with it. The lib just wraps the existing user program, handling subscription changes after user init/update. Its [sole dependency](https://github.com/kspeakman/ElmishSubCliDemo/blob/main/src/Elmish.LiveSubs/Elmish.LiveSubs.fsproj#L15) is on Elmish. I also changed its implementation to be as similar as possible to the one in this PR.

The `Time.every` subscription [here](https://github.com/kspeakman/ElmishSubCliDemo/blob/main/src/Demo/App.fs#L13-L22) can be used for testing. It uses Timer instead of JS interval.

src/sub.fs
kspeakman: Removed

TysonMN:
> I'm not familiar with using Elmish from the console.

Oh, sorry. I assumed that it would be easy for you to create a CLI integration.

> If we're ultimately talking about Elmish.Wpf, the Elmish.LiveSubs project should be directly usable with it.

If you think this is easy, then can you fork Elmish.WPF on GitHub and modify one of the samples that uses subscriptions to use your new API (then then share a link to it here)?

et1975 (approved):
I think it's good enough now to merge, release a beta and start checking what breaks downstream.

TysonMN:
> I think it's good enough now to merge, release a beta and start checking what breaks downstream.

FWIW, I am still not convinced that these features require the user to provide an ID.

kspeakman:
> Oh, sorry. I assumed that it would be easy for you to create a CLI integration.

You'd think it would be easy. But it turns out not. Especially managing simultaneous / independent input and output to the console. It needs an integration piece that adapts Elmish to console.

> If you think this is easy, then can you fork Elmish.WPF on GitHub and modify one of the samples that uses subscriptions to use your new API (then then share a link to it here)?

I'll give it a try.

> FWIW, I am still not convinced that these features require the user to provide an ID.

If you look at [this](https://github.com/kspeakman/ElmishSubCliDemo/blob/main/src/Demo/App.fs#L64), the user doesn't have to. Further explained [here](https://github.com/elmish/elmish/pull/248#issuecomment-1233171094).

src/program.fs
```diff
@@ -45,13 +45,14 @@ module Program =
           update = fun msg -> update msg >> fun state -> state, Cmd.none
           view = view
           setState = fun model -> view model >> ignore
-          subscribe = fun _ -> Cmd.none
+          subscribe = fun _ -> []
```
et1975: `Sub.none`?

src/program.fs
```diff
@@ -12,7 +12,7 @@ namespace Elmish
 type Program<'arg, 'model, 'msg, 'view> = private {
     init : 'arg -> 'model * Cmd<'msg>
     update : 'msg -> 'model -> 'model * Cmd<'msg>
-    subscribe : 'model -> Cmd<'msg>
+    subscribe : 'model -> Sub<'msg> list
```
et1975: I wonder if we should attempt the symmetry with `Cmd`  - list being an impl detail?

et1975 (approved):

src/program.fs
kspeakman: I always felt cognitive dissonance and confusion with Cmd in Elm. It's name is singular but it represents multiple items. But then sometimes it behaves singular, like `Task.perform` resulting in a Cmd. One of the first reasons I looked at Elm source code was because this felt too magical.

In Elmish, the Cmd type alias always felt in the way, unnecessarily hiding the list. If it wasn't such a big change, the best thing IMO would be to drop the list alias and rename Effect to Cmd to bring it in line with Sub.

src/program.fs
kspeakman: Fixed

src/program.fs
kspeakman: If we make Sub like Cmd, we need to come up with another name for a single Sub item. I suggest `Svc<'msg>` since it's essentially a service with start and stop behavior. In JS, that's mainly about listening to DOM events. But in dotnet, it could be a long-running Task or Async on the thread pool.

et1975: Maybe we can keep the individual sub type anonymous? Individual sub doesn't appear in any public signature, and even internal impl seems to only use the lists.
```
type Sub<'msg> = {| SubId: SubId; Start: Dispatch<'msg> -> IDisposable |} list
```

kspeakman: For public signature, there are two audiences for the Sub type. One is library authors. Another is end users who need to create their own subscriptions. (The most likely scenario until libraries emerge.)

Anonymous record seems problematic here. A library/user fn that needs to adjust a Sub item in some way would have to repeat the record definition in a type annotation. And it lacks some record features, namely destructuring. If we're going for parity, this would also be different from what we do with Cmd, where an individual item has a type definition, even if it's an alias. (To eliminate type aliases, a more readable choice might be to define a record wrapper for the fn.)

kspeakman:
@TysonMN The Elmish.LiveSubs lib doesn't seem compatible with Elmish.Wpf. LiveSubs wraps the user's Elmish program (Model and Msg) in its own, similar to other Elmish libs. I could not find an existing way to do this in Elmish.Wpf, e.g. `WpfProgram.map`.

I looked at adding a WpfProgram fn, but mapping the bindings was a problem. The existing `BindingData.mapModel` seems to be for a different purpose that is not compatible with this usage. Maybe I missed a more appropriate method?

This PR will not have the same issue, since the subs are integrated into the Elmish loop instead of wrapping the user program.

As mentioned above, the sub id does not have to be a user-exposed concern. The sub will probably provide something by default anyway -- config settings that should cause a restart when changed. Like timer interval. The user or other libs can add to the ID only if needed.

TysonMN:
> LiveSubs wraps the user's Elmish program (Model and Msg) in its own, similar to other Elmish libs.

Can you link to an example of such a library?

kspeakman:
> > LiveSubs wraps the user's Elmish program (Model and Msg) in its own, similar to other Elmish libs.
> 
> Can you link to an example of such a library?

Elmish.HMR wraps Model and Msg
https://github.com/elmish/hmr/blob/v3.x/src/hmr.fs#L44-L50

Elmish.Browser wraps Msg
https://github.com/elmish/browser/blob/v3.x/src/navigation.fs#L18-L20

kspeakman:
That and parent/child is a common pattern in MVU.

TysonMN:
Don't let the public API hold you back. In your demo, you can make anything public that you want.

> I looked at adding a WpfProgram fn, but mapping the bindings was a problem. The existing `BindingData.mapModel` seems to be for a different purpose that is not compatible with this usage. Maybe I missed a more appropriate method?

I think you are looking for `Bindings.mapModel` and `Bindings.mapMsg`.

et1975:
Released as 4.0.0-beta-5, thank you @kspeakman!

kerams:
Just to confirm, when upgrading every usage of `Cmd.ofSub` should simply be replaced with `Cmd.ofEffect`?

kspeakman:
@kerams Yes. v3 Sub = v4 Effect. It's a name change only.

kspeakman:
I will look at adding some documentation and/or a migration guide.

PR #249, opened 2022-09-09
Update basics.fsx

helegrod:
When going to run the code, I noticed that the state actually varied from -1 to 4. Alternatively the docs could be changed.


et1975:
Thanks!

PR #255, opened 2022-10-25
Fix doc build

kspeakman:
The last commit to fix trailing whitespace failed. Build / package step complained about remark-parse being missing. GenerateDocs works again after I installed it (npm package).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,8 @@
         "": {
             "name": "hmr",
             "dependencies": {
-                "nacara": "^1.5.1"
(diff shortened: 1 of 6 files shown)


et1975 (approved):

MangelMaxime:
@kspeakman His `remark-parse` used for something?

It looks like it has been added has a dependency but it doesn't seems to be used.

kspeakman:
I was very confused by this as well. It was nothing I changed. I think this dependency was added in #251. Its build step also failed.

PR #257, opened 2022-10-30
Dispatch queues during startup

### LINKED ISSUE #256: Double subscribe when dispatching during sub startup
### Description

When a message is dispatched during subscription startup, the subscription will start twice. So there are two running instances when there should be one.

### Repro code

```fsharp
// from Elmish.HMR PR I'm working on
module Sub =

    let tick intervalMs msg onStart=
        let subId = ["tick"]
        let start dispatch =
            let intervalId = JS.setInterval (fun () -> dispatch msg) intervalMs
            dispatch onStart // double subscribe caused here
            { new System.IDisposable with
                member _.Dispose() = JS.clearInterval(intervalId) }
        subId, start

let private subscribe model =
    [Sub.tick 1000 Tick Reloaded]
```

If I put the dispatch in a 1ms `setTimeout`, there will only be one subscribe. _In this example, `init` and `update` never return any cmds._

I think I figured out why this happens. In program startup, dispatch is in run mode. So the subscription starts and enters the dispatch loop. But I haven't set the `activeSubs` yet. I can't be sure if a sub is active until it starts successfully. So in the dispatch code, the subs are recalculated again after user updates, which triggers another subscription start. Here's what I think the filtered call stack looks like.

```bash
Program.run
├── user init
├── Subs.Fx.change (activeSubs = [])
│   └── SUBCRIPTION STARTS, dispatches a msg
│       ├── Program.runWithDispatch.dispatch (not reentered, runs)
│       │   ├── user update
│       │   ├── Subs.Fx.change (activeSubs = [])
│       │   │   └── SUBCRIPTION STARTS, dispatches a msg
│       │   │       └── Program.runWithDispatch.dispatch (reentered, queues)
│       │   └── activeSubs <- ["tick", IDisposable]
│       └── Program.runWithDispatch.dispatch (loops next message)
│           ├── user update
│           ├── Subs.Fx.change (no changes, nothing starts)
│           └── activeSubs <- ["tick", IDisposable] (same)
└── activeSubs <- ["tick", IDisposable]
```

### Possible solutions

We could speculatively set `activeSubs` with the subIds and empty IDisposables before trying to start them. When dispatch gets called, it won't try to start subs again. However, we won't know which subs are really active until they start successfully.

Even with that, `subscribe` will get called more than once on program startup. There could be other bugs arising from this complexity. It seems more of a problem that the dispatch loop can run during program startup. Msgs should probably be queued until startup is over, then looped thru. Unless I've completely misread the situation.

### Expected and actual results

Expected: subscription is only started once.
Actual: subscription is started twice, causing 2 Tick msgs per second

### Related information 

* elmish version: **4.0.0-beta-5**
* fable-compiler version: **3.4.9**
* fable-core version: **3.7.1**
* Operating system: **Windows 10 21H2**

kspeakman:
resolves #256

I'm throwing this out there as a possible solution. Feel free to reject if it's inappropriate or there's a better way.

The logic from dispatch is all there and the same, but I moved message processing to its own function. The idea being that the queue (ring buffer) is the dividing line between the user dispatch and message processing.

- `processMsgs` does what its name suggests. It uses the queue as its msg source.
- `dispatch` only queues user msgs. If `processMsgs` is not running (reentered = false), it will be started.

During program startup, reentered = true is set to make dispatch queue, without starting `processMsgs`. That way the user init and subscription changes don't get off in the weeds processing updates before both are finished. After those are done, `processMsgs` is called, kicking off the update loop.

I think this was not an issue before v4 subscriptions because there were only Cmds to dispatch msgs. `subscribe` was like a special purpose `init`. `subscribe` and `init` cmds were batched together and executed as the last startup action. So there wasn't a need to queue during startup.

### DIFF
--- a/src/program.fs
+++ b/src/program.fs
@@ -167,33 +167,36 @@ module Program =
         let mutable activeSubs = Subs.empty
         let mutable terminated = false
         let rec dispatch msg =
-            if terminated then ()
-            else
-                if reentered then
-                    rb.Push msg
-                else
+            if not terminated then
+                rb.Push msg
+                if not reentered then
                     reentered <- true
-                    let mutable nextMsg = Some msg
-                    while not terminated && Option.isSome nextMsg do
-                        let msg = nextMsg.Value
-                        if toTerminate msg then
-                            Subs.Fx.stop program.onError activeSubs
-                            terminate state
-                            terminated <- true
-                        else
-                            let (model',cmd') = program.update msg state
-                            let sub' = program.subscribe model'
-                            program.setState model' dispatch'
-                            cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
-                            state <- model'
-                            activeSubs <- Subs.diff activeSubs sub' |> Subs.Fx.change program.onError dispatch'
-                            nextMsg <- rb.Pop()
+                    processMsgs ()
                     reentered <- false
         and dispatch' = syncDispatch dispatch // serialized dispatch
+        and processMsgs () =
+            let mutable nextMsg = rb.Pop()
+            while not terminated && Option.isSome nextMsg do
+                let msg = nextMsg.Value
+                if toTerminate msg then
+                    Subs.Fx.stop program.onError activeSubs
+                    terminate state
+                    terminated <- true
+                else
+                    let (model',cmd') = program.update msg state
+                    let sub' = program.subscribe model'
+                    program.setState model' dispatch'
+                    cmd' |> Cmd.exec (fun ex -> program.onError (sprintf "Error handling the message: %A" msg, ex)) dispatch'
+                    state <- model'
+                    activeSubs <- Subs.diff activeSubs sub' |> Subs.Fx.change program.onError dispatch'
+                    nextMsg <- rb.Pop()
 
+        reentered <- true
         program.setState model dispatch'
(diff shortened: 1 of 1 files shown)


kspeakman:
There's a fair question: can subscription start return a Cmd instead of immediately executing.

That was my [initial implementation](https://github.com/kspeakman/ElmishLiveSubsDemo/blob/main/src/Elmish.LiveSubs/Program.fs) to demo the possibility. It adds a lot of complication.

When start occurs in Cmds, there are things to add. Because Cmd returns unit, I can't immediately know that the sub started successfully, nor how to stop it (IDisposable in v4). Until I find out, I need to track whether it's starting, running, or stopping. So I don't accidentally start or stop it more than once.

How does the program gets notified of a successful start, and the stop fn? The mechanism provided by Cmd is msg dispatch. That adds a lot of complication to the program's main loop -- needing to wrap the user's msgs to also handle sub msgs. Alternatively, we could directly call an internal fn instead of dispatching, like `subNotify`. But then we have multi-threading problems on non-JS platforms. We might have to expose another sync function or change syncDispatch to handle both.

Whereas this PR adds to program init the exact same mechanism already used by update. And for the same reason -- so that only one copy of the decision loop is running. The reason it feels awkward is because `init` was originally designed as both a constructor fn and the first decision loop. With subsequent decision loops moving to `update`. So the same program logic has to be duplicated for both.

et1975 (approved):

PR #260, opened 2022-10-31
WIP subscription v4 docs

### LINKED ISSUE #259: Subscription doc updates for v4
Subscription docs need to be updated to reflect v4 usage.

kspeakman:
resolves #259

TODO
- [x] Convert subscription docs to use new subs
- [x] v3 users need docs too, create a page for them
- [x] Add a v3 to v4 migration section
- [x] Advanced subscription examples (ID related)
  - [x] Multiple timers on same page
  - [x] Timer that can change intervals
- [x] Test that the doc build works 🙃


docs/docs/subscription.fsx
```diff
@@ -23,61 +23,141 @@ We can setup forwarding of those events to be processed by our `update` function
...
-|> Program.run
+A subscription has an ID, `["timer"]` here, and a start function. You're probably wondering why the ID is a list.
+It allows you to include dependencies. Later we will use this to make the timer change intervals.
 
+Here, the timer subscription is always returned, so it will stay running as long as the program is running.
+
+Let's look at an example where the timer can be turned off. First we add the field `enabled`.
+*)
+
+module Example2 =
```
MangelMaxime: @kspeakman If you want to avoid the user seeing `module Example2 =` in the ouput.

You can use `(*** hide ***)` and `(*** show ***)` to hide portion of the F# code.

*Note: `(*** hide ***) will hide the code up to the next `(**` comments.*

Example:

```fs
(*** hide ***)
let a = 42 // This will not be in the output 

(*** show ***)
let b = 42 // This will be in the ouput
```


```fs
(*** hide ***)
let a = 42 // This will not be in the output 

(**
This text will be in the output and the next snippet too
*)
let b = 42 // This will be in the ouput
```

kspeakman: Thanks for the feedback. I made some changes based on it. I tried to make the module names match the example's purpose. For one-off notes, I hid the module name to make it appear as if it was in the previous module.

kspeakman:
I'd love to get some feedback on the documentation content so far.

* Is there something obvious that I didn't cover?
* How did you feel about the sections on subscription reuse and ID usage?
* Does it flow from one topic to the next well enough?

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+        match msg with
+        | Tick current ->
+            { model with
+                current = current
+            }
+
 (**
 
-Note that "simple" is not a requirement and is just a matter of convenience for the purpose of the example!
+Note that "simple" is not a requirement and is just a matter of convenience for the example!
```
MangelMaxime: Wrongly place note?

I suppose that `simpe` refers to `mkSimple` and so it should we be place after seeing it in the code?

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+                    (fun _ -> dispatch (onTick DateTime.Now))
+                    1000
+            { new IDisposable with
+                member _.Dispose() = JS.clearInterval intervalId }
+        start
+
+    let subscribe model =
+        [ ["timer"], timer Tick ]
+
+    Program.mkSimple init update (fun model _ -> printf "%A\n" model)
```
MangelMaxime: Can we avoid using `mkSimple`?

When I started learning Elmish/Elm, it was always confusing how to move from "simple" program to "normal" program.

IHMO, `mkSimple` should not exist because you can't really build a "real" application with it.

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+    let subscribe model =
+        [ if model.enabled then
+            ["timer"], timer Tick ]
+
+    Program.mkSimple init update (fun model _ -> printf "%A\n" model)
+    |> Program.withSubscription subscribe
+    |> Program.run
+
+(**
+All that's left is to add user interaction to change `enabled`. The timer will stop or start accordingly.
```
MangelMaxime: > All that's left is to add user interaction to change `enabled`

I don't really like the "All that's left" because it seems to imply this is simple to do but if the user don't know how to do it he could be frustrated.

MangelMaxime: I also think it would be better to show how to add that user interaction. 

We can using the React DSL for that and adding a mention that user is not limited to HTML interaction and here we just use it for simplicity.

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+
+You probably noticed that `subscribe` returns a list. So let's add another timer that's always on.
```
MangelMaxime: I would re word:

Notice how `subscribe` returns a list. This is so, you can add several subscription at the same time, etc.

The change is because if the user didn't notice it, it could be miss interpreted by him.

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+(*** show ***)
+    let subscribe model =
+        [
+            if model.enabled then
+                ["user"; "timer"], timer Tick
+            ["page"; "timer"], timer Tick
+        ]
+
+(**
+Subscription IDs need to be unique. It's not necessary to use two ID segments... I just liked it better than camel casing the words.
```
MangelMaxime: I would remove the personal note and just mentioned that user can use:

- `[ "users"; "timer" ]`
- `[ "usersTimer" ]`

depending on where there preference leans to.

docs/docs/subscription.fsx
```diff
@@ -23,61 +23,162 @@ We can setup forwarding of those events to be processed by our `update` function
...
+
+Small stylistic note: In the timer subscriptions above, I used `let` to define the `start` fn, then return it. You will often see a function return value written like this:
```
MangelMaxime: This feels a bit confuse, because you are referring to a snippet above but then user needs to refer to the snippet below.

I think just referring to the one below is good enough or move the remark above and not have 2 snippets.

docs/docs/subscription.fsx
```diff
@@ -160,12 +261,195 @@ module App =
...
 
     Program.mkSimple init update (fun model _ -> printf "%A\n" model)
-    |> Program.withSubscription subscription
+    |> Program.withSubscription subscribe
     |> Program.run
+
+(**
+Notice `Sub.map` takes an id prefix as its first parameter. This helps keep subscriptions distinct from each other.
+
+Before Sub.map, Second and Hour have the same ID: `["timer"]`. After Sub.map, their IDs are: `["hour"; "timer"]` and `["second"; "timer"]`.
```
MangelMaxime: "Sub.map" should be between code statement:

```
`Sub.map`
```

2 places to change

docs/docs/subscription.fsx
```diff
@@ -160,12 +261,195 @@ module App =
...
+        {
+            current = DateTime.MinValue
+            intervalMs = 1000
+        }
+        , Cmd.ofEffect (Time.now Tick)
+
+(**
+### IDs and dependencies
+
+I mentioned earlier that ID is a list so that you can add dependencies to it. What does this mean?
```
MangelMaxime: I would rephrase to:

Earlier, we saw that ID is a list ... It means ...

docs/docs/subscription.fsx
```diff
@@ -160,12 +261,195 @@ module App =
...
+module IdDeps =
+
+    open EffectTimer // reuse last example's code
+
+(*** show ***)
+    let subscribe model =
+        [ ["timer"; string model.intervalMs], Sub.timer model.intervalMs (Time.now Tick) ]
+
+(**
+Now that intervalMs is part of the ID, an interesting thing happens. When the interval changes, the timer will stop then restart with the new interval.
```
MangelMaxime: I would remove "an interesting thing happens"

MangelMaxime (commented):
I made a pass to review the documentation. Globally, I think the different step shown makes sense and are in the right order.

I am just not a fan of the phrasing used in some places:
- The interrogation used in some places
- The use of `I` sentences
- Asking the user, if he noticed something or not

Because, it feels like the reader is being judged a bit :). 

---

Nacara supports [block container](https://mangelmaxime.github.io/Nacara/nacara-layout-standard/markdown-features.html#block-containers)

Perhaps, it could be used in some places when you say "Note: ..."

kspeakman:
@MangelMaxime Thank you for the excellent feedback. I will work on the things you mentioned.

docs/docs/subscription.fsx
kspeakman: I noticed that existing examples from v3 printed to console and lacked user interaction. So I thought maybe the .fsx file was meant to stay locally runnable. If that's not a goal, I can add React DSL examples, no problem.

docs/docs/subscription.fsx
kspeakman: Do you think I should just leave it out and convert all the subscriptions to return `fun dispatch -> . . .`?

When I first started making my own effects (v3 subs), I saw the anonymous fn syntax used everywhere in Elmish code. But for some reason this just didn't fit in my brain. Using `let f dispatch = . . .; f` made it readable for me, but maybe is just me.

docs/docs/subscription.fsx
kspeakman: It is only there to avoid multiple conditional phrases in a row. "Now that ..., when the ..." It's like code with nested if statements. Takes extra effort to understand. Terminating the first conditional phrase helps. Changing to a statement of fact: "Now `intervalMs` is part of the ID." is stating an obvious fact without making a direct connection to the next sentence. I left the sentence in as more of a transition into advanced ID usage. Then the next section dives even deeper. Writing can be quite a logic puzzle sometimes. Basically writing code for human brains. I'll try to find a different way to phrase it.

docs/docs/subscription.fsx
kspeakman: I'm dense sometimes. I didn't quite understand what the change should be. I do understand that there are multiple mentions of Sub.map and only one of them is code-formatted. Probably just a misguided personal preference on my part, to avoid overemphasizing Sub.map when repeating it. Maybe a sign that I used it too many times. I'll change it to whatever you prefer.

docs/docs/subscription.fsx
kspeakman: Will convert to regular program and remove this note.

docs/docs/subscription.fsx
kspeakman: (I think it's because Cmd/Sub feels like a data type in its usage, but has a partial-application-based function API for its implementation. The asymmetry didn't compute for me.)

docs/docs/subscription.fsx
MangelMaxime: Personally, my main usage of the fsx is to have intellisense/compiler assistance so that I don't make typo in the API used.

Printing to console was how Eugene wrote the initial documentation, but personally I always found it difficult to understand because interactivity is limited with console. In general, I think HTML/React is "easy" enough to understand for our usage here.

docs/docs/subscription.fsx
MangelMaxime: Could you please show me the 2 versions of the code? I am a bit lost here.

docs/docs/subscription.fsx
MangelMaxime: Personally, when I refer to some stuff which are from an API, or can be consider code I always use decorated them with ` as I find it user to say, ok this is some code blabla.

AlbertoDePena:
Forgive my ignorance but I would love to read over the documentation... Just don't know how... Is it something I can build on machine? I love Elmish and use it (in it's purest form) at work for our new projects. Anything I can do to help just let me know :)

MangelMaxime:
@AlbertoDePena Yes, you can build the documentation locally:

1. Clone the repo
2. `npm install`
3. `dotnet tool restore`
4. `dotnet fake build -t WatchDocs`
5. The docs should be served now

docs/docs/subscription.fsx
kspeakman: Here's the style I used in the docs:

```fsharp
let timer onTick =
    let start dispatch = // name fn
        . . .
    start // return named fn
```

Here's the style Elmish libraries often use:

```fsharp
let timer onTick =
    fun dispatch -> // return anonymous fn
        . . .
```

I found this very confusing. F# functions are curried anyway, so the anonymous fn return is equivalent to making dispatch a regular parameter:

```fsharp
let timer onTick dispatch =
    . . .
```

Eventually I understood that the anonymous fn return is trying to tell the user to only provide the first parameter and a function should be returned. To make the point even more obvious in the docs (and my code), I explicitly define a named `start` function and return it.

But maybe that's just me. Or perhaps it should use the anonymous function return style to be consistent with Elmish libraries' source code.

docs/docs/subscription.fsx
kspeakman: Sounds good. I'll say at the top that the examples use Fable.Elmish.React, and link to its doc. So they have a starting place for that.

docs/docs/subscription.fsx
MangelMaxime: I see, I don't have a real preference.

In general, I used the anonymous still because I was just mimicking what Eugene wrote.

docs/docs/subscription.fsx
kspeakman: I see why the original example was mkSimple. When you get to the section with nested subs, it adds a lot of noise also demonstrating parent-child msg mapping.

docs/docs/subscription.fsx
kspeakman: I ended up just including `open Feliz` above the `view` function. So it's clear what is being used.

docs/docs/subscription.fsx
kspeakman: I added a note at the very bottom of the subscription page to demonstrate the difference. Otherwise I kept the named fn approach for more explicitness.

docs/docs/subscription.fsx
kspeakman: I ended up removing the example. The point that subscribe returns multiple subs is already implicit in that it returns a list. And is also demonstrated in later examples.

kspeakman:
@MangelMaxime @et1975 This is ready for final review. If there's anything major that needs adding, we'll probably want to do it as a separate issue to not hold up the v4 release. I'll be watching this PR closely so I can turn around any minor issues quickly.

et1975 (approved):

Issue #263, opened 2022-12-12
Elmish for WebSharper apps


granicz:
I was looking into compiling Elmish with WebSharper and it seems pretty straightforward. What is the recommended way to contribute that back to this repo?

et1975:
If it's just a project/settings sort of thing a PR into v4 branch.
If it introduces rendering specifics, then it would need a separate repo,
like we have for "react".
If it needs certain platform support, then it's more like "browser".

granicz:
Cool, thanks - no rendering specifics or extra platform code, just enabling Elmish to be incorporated into WebSharper apps - so a PR into v4 sounds like the way to go. It would be an extra project file that references the existing `.fs` files from `src` and one additional `.fs` file. I can also plug in creating a new `WebSharper.Elmish` package along with the current `Elmish` and `Fable.Elmish` ones.

Given that Fable.Elmish is baked into `src` files via `#IF FABLE_COMPILER` directives, would it make sense to add the above under `src/WebSharper` or under `netstandard/WebSharper`? The latter might work better.

et1975:
Thanks Adam, released as 4.0.0-beta-6.

granicz:
Thanks @et1975 !

PR #264, opened 2022-12-13
Elmish for WebSharper apps

granicz:
Added #263

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ v2.0 and above releases use `dotnet` SDK and can be installed with `dotnet nuget
 For use in a Fable project:
 `paket add nuget Fable.Elmish -i`
 
+For use in a WebSharper project:
+`paket add nuget WebSharper.Elmish -i`
+
 If targeting CLR, please use Elmish package:
 `paket add nuget Elmish -i`
 
--- a/build.fsx
+++ b/build.fsx
@@ -27,12 +27,15 @@ open System.IO
 let projects  =
     !! "src/**.fsproj"
     ++ "netstandard/**.fsproj"
+    ++ "netstandard/websharper/**.fsproj"
 
 Target.create "Clean" (fun _ ->
     Shell.cleanDir "src/obj"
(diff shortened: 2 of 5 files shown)


et1975 (approved):

kspeakman:
@granicz I'm embarrassed to say I'm a bit lost on this feature. My understanding certainly isn't required to proceed. But if you get a moment, can you explain the context around this?

granicz:
Sure, let me take a bird's eye view to illustrate. While Elmish is built on .NET, it is a particularly useful abstraction to structure UI applications of all sorts (desktop, mobile, web, etc.). It by design does not involve or care about any given view, but you can build further logic around it to target different "UI platforms" such as WPF, JavaScript/DOM, Xamarin, etc. - and instantiate Elmish's abstract notion of a "view" with something concrete in those.

Now, applications that end up running outside of .NET (say, in JavaScript) need the Elmish codebase processed/transpiled/translated to their own target language/representation. Transpiling often reinterprets the underlying code with additional or different semantics, either via compiler directives you see for Fable (`#if FABLE_COMPILER ... #else ...`), or via other metaprogramming techniques (such as code attributes, shadowing of identifiers, or full-blown proxies such as in WebSharper). For this reason, one could argue they should be placed in the source repository - which is a matter of taste and/or necessity of the underlying toolset.

This repository has always provided the translation to JavaScript with Fable, allowing Fable apps to use and build on Elmish, but not with WebSharper, which is the hole I am filling with the above PR. Once merged, it will enable WebSharper apps to use and build on Elmish in the same way as well.

Hope this helps. Let me know if you need further clarification, happy to elaborate further.

et1975:
Had to flatten it a bit as nuget props wheren't happy with the relative paths.

kspeakman:
@granicz Thanks for the explanation. Makes sense. I just wasn't aware of WebSharper's mechanism of usage. The code added for WebSharper.Elmish is amazingly small, so it didn't give me any hints. :)

PR #271, opened 2023-02-10
Update subscription.fsx

### LINKED ISSUE #270: Docs: String.replicate piece raises an exception
In particular in: https://elmish.github.io/elmish/docs/subscription.html#ids-and-dependencies

```fs
                                        Html.div [
                                            prop.key (string now.Ticks)
                                            let timestampStr =
                                                let s = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")
                                                let zeros = 23 - s.Length
                                                s + String.replicate zeros "0"
                                            prop.text timestampStr
                                        ]
```

As `now.ToString("yyyy-MM-dd HH:mm:ss.ffff")` is of length 24, running this would give out errors in the browser console. Best fix would be to do it like:

```fs
                                        Html.div [
                                            prop.key (string now.Ticks)
                                            let timestampStr = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")
                                            prop.text timestampStr
                                        ]
```

I tried making an edit myself, but I am confused as to where the main site docs are supposed to be in the latest branch. Also, I am sure you are aware of this, but all the example runners are missing `|> Program.withReactBatched "app"`. Yesterday I had a lot of trouble figuring out that I also need to add a `<div id="app">` element `index.html`'s body.

mrakgr:
Fixes #270.

### DIFF
--- a/docs/docs/subscription.fsx
+++ b/docs/docs/subscription.fsx
@@ -702,10 +702,7 @@ Let's finish things off with view functions.
                                     Html.td [
                                         Html.div [
                                             prop.key (string now.Ticks)
-                                            let timestampStr =
-                                                let s = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")
-                                                let zeros = 23 - s.Length
-                                                s + String.replicate zeros "0"
+                                            let timestampStr = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")
                                             prop.text timestampStr
                                         ]
                                     ]


kspeakman:
Why not change 23 -> 24? The zeros have a purpose. The format "ffff" sometimes returns less than 4 digits. This causes the displayed timestamp to shift. Which is jarring and hard to read on faster timers.

mrakgr:
@kspeakman Does that really happen?

```fs
                                            let now = DateTime.Parse("2023-02-13 18:33:01")
                                            let timestampStr = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")
```

I am playing around, trying to get it to omit the trailing 0s, but I can't do it.  Parsing the above date just outputs: `2023-02-13 18:33:01.0000`. The same goes if I have it be `let timestampStr = now.ToString("yyyy-MM-dd HH:mm:ss.ffff")`. In that case it outputs `0001-01-01 01:22:00.0000`.

kspeakman:
> Does that really happen?

Yes, that's why I stated it and why the zero padding is there. Did you try to run it inside a Fable Elmish app, or in F# interactive? One is a JS Date object. The other is a .NET DateTime object. Their behaviors differ.

mrakgr:
I ran it as a Fable Elmish app inside the browser. Forgive me, I am skeptical that what you are saying is true. Just the format itself `ffff` specifies that there should be 4 numbers at the end. I tried to verify you claim that it would get cut off, and that doesn't happen when I use it on DateTime objects where I'd expect it would happen. Lastly, the original code throws an exception 100% of the time, meaning it hasn't been tested even once before being put out.

In order to quell my skepticism, could you show me on what kind of input the trailing zeros get cut?

kspeakman:
> Forgive me, I am skeptical that what you are saying is true.

Wow, aggressive. Not a good way to interact with open source projects.

I started with a format letter code. That could be the source of the inconsistent length that I remember dealing with, rather than "ffff". And why the zero padding got added. Eventually I switched to the format string. But must have forgotten about the zero padding. String replicate didn't crash on negative indexes until recently.

That considered, the PR looks fine. Next time, please try assuming everyone is working in good faith.

MangelMaxime:
Thank you @kspeakman for looking into this PR and explaining why it was not failing in the past. Sometimes, bug are coming from subtle changes not related to the project itself 😅

@mrakgr Thanks for catching the error.

mrakgr:
My bad. I guess it was tested, but got changed, sorry about that.

kspeakman:
No worries. I've done far worse, learned the hard way. 🙃

PR #280, opened 2023-10-16
reverse order of cmds and subs evaluation

et1975:
Fix for [browser/#66](https://github.com/elmish/browser/issues/66)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+## 4.1.0
+* Reverse order of subs and cmds evaluation [#66](https://github.com/elmish/browser/issues/66)
+
 ## 4.0.2
 * Loosen the tasks contraint when we don't care about the result (#277).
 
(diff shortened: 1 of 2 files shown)


kspeakman:
I think this change makes sense. In effect, it sets up subscriptions before the user init instead of after.

Be aware that this is potentially a breaking change. If the user currently depends on msgs emitted from `init` Cmds to occur before msgs from subscriptions. A specific scenario doesn't come to mind. But I also didn't think of using `newUrl` from top level `init`.

kspeakman (commented):
Left comment in issue.

MangelMaxime:
On my side, the question I have is do we need reverse the order only when initialising the program or also in the inner loop (like done in this PR) ?

I don't have a scenario in mind for that, but I because I can't shake off the question of my head I preferred to asked it out loud 😅

kspeakman:
I think consistency is best. And in general, this is exactly the kind of scenario I had in mind. Subscriptions setup services used by the page.

The only lingering doubt I have is: what if I want to run something before subscriptions? I guess the answer there is to run it before running the program or as the first subscription.

kspeakman:
Forgot the obvious: running something before subscriptions is already in the page's control. It can not return any subscriptions, run whatever initialization it wants, then return them.

PR #299, opened 2025-03-18
fix: capture error when generating "tasks" + add CmdTests suits

MangelMaxime:
This PR fixes a bug where, if you have an exception happening before entering the `async` code then the exception would not be captured by the command.

### DIFF
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -84,12 +84,8 @@ module Cmd =
             let bind dispatch =
                 async {
                     try
-                        let! r = task arg |> Async.Catch
-                        dispatch (
-                            match r with
-                            | Choice1Of2 x -> ofSuccess x
-                            | Choice2Of2 x -> ofError x
-                        )
+                        let! r = task arg
+                        dispatch (ofSuccess r)
                     with x -> dispatch (ofError x)
                 }
             [bind >> start]
@@ -102,10 +98,8 @@ module Cmd =
             let bind dispatch =
                 async {
                     try
-                        let! r = task arg |> Async.Catch
-                        match r with
-                        | Choice1Of2 x -> dispatch (ofSuccess x)
-                        | _ -> ()
+                        let! r = task arg
+                        dispatch (ofSuccess r)
                     with _ -> ()
                 }
             [bind >> start]
@@ -118,10 +112,8 @@ module Cmd =
             let bind dispatch =
                 async {
                     try
-                        let! r = task arg |> Async.Catch
-                        match r with
-                        | Choice2Of2 x -> dispatch (ofError x)
-                        | _ -> ()
+                        let! _ = task arg
+                        ()
                     with x -> dispatch (ofError x)
                 }
             [bind >> start]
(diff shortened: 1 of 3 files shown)


src/cmd.fs
```diff
@@ -173,30 +181,36 @@ module Cmd =
...
-                (task arg)
-                    .``then``(ofSuccess >> dispatch)
-                    .catch(unbox >> ofError >> dispatch)
-                    |> ignore
+                try
+                    (task arg)
+                        .``then``(ofSuccess >> dispatch)
+                        .catch(unbox >> ofError >> dispatch)
+                        |> ignore
+                with x -> x |> unbox |> ofError |> dispatch
```
MangelMaxime, line 181: Here I followed the pattern from the existing implementation of using `unbox`.

This seems to be needed because of the `#exn` condition, when others Cmd (outside of `Promise`) don't have that same requirements they only rely on `exn` and not `#exn`.

MangelMaxime:
@et1975 The CI is once again broken by Fake

What solution would you prefer for fixing the CI?

In my other projects, I removed FAKE completely in favor of a pure F# project written as a[ CLI application](https://github.com/glutinum-org/cli/tree/main/src/Glutinum.Build) or by using [Fun.Build](https://github.com/slaveOftime/Fun.Build) which is kind of like Fake but didn't break on me (yet?).

We could keep FAKE and pinned the version of the dependencies but it seems like FAKE don't work with .NET9 so we would probably have the same issue later when trying to upgrade to a new .NET version.

et1975:
I find using fake as a library is not as fragile as using fake as .net
tool. I'm running the scripts on .net 9 without issues at work. Just need
to fix it.


Regarding the problem of exceptions... Do we even need Async.Catch then?

et1975:
Pinned the dependencies and it seems to build OK atm, except that moving up to v9 SDK had to be postponed - WebSharper project fails to build.
cc  @granicz

granicz:
@et1975 The WebSharper project uses WS6 - which is quite old, and WS8 is in the release pipeline for next week, but what's the error exactly?

MangelMaxime:
> I find using fake as a library is not as fragile as using fake as .net tool. I'm running the scripts on .net 9 without issues at work. Just need to fix it.

If you know how to fix it can I let you do it?

> Regarding the problem of exceptions... Do we even need Async.Catch then?

It seems like we don't need it anymore, I removed the redundant error catching from `promise` code too.

et1975 (approved):

et1975:
rabased on v4.x latest to fix the build

et1975:
Released as 4.3.0

MangelMaxime:
Thank you :)

et1975:
> Thank you :)

Thanks for fixing it. When I wrote that I couldn't imagine anyone doing anything outside of the computation they are returning, but this is safer.

MangelMaxime:
Me neither, the way the issue was detected was because of Fable.Remoting usage with an invalid data given to it (due to interop or `unbox` called somewhere to by pass the type system).

Outside of that, I think the issue could have gone unseen for even longer 🤣

PR #300, opened 2025-03-20
feat: replace `Cmd.OfAsync.start` with `AsyncHelpers.start`

### LINKED ISSUE #298: Should `start`  function at `Cmd.OfAsync.start` be moved somewhere else?
Currently,

The API exposed by Elmish from 

https://github.com/elmish/elmish/blob/946e1bc6eb67d3e10793d5802c199ed21d930b77/src/cmd.fs#L121-L126

Allow the user to access `Cmd.OfAsync.start`

This feels wrong to me because `Cmd.OfAsync.start` is not a command related function, but instead a helper function to simulate the same behavior between Fable and .NET.

I think we should move it inside of the `Prelude.fs` file and put it inside a `AsyncHelpers` or `AsyncProxy` module. I am not using `Async.start` here because of the conflict with the original `Async` API.

If we were to mark the API as `internal` we could use `Async` for the module name, but I think it is still useful to expose the helpers for people we write their own `Cmd` functions

MangelMaxime:
Fix #298

### DIFF
--- a/src/prelude.fs
+++ b/src/prelude.fs
@@ -33,3 +33,10 @@ module internal Timer =
         t.Enabled <- true
         t.Start()
 #endif
+
+module AsyncHelpers =
+#if FABLE_COMPILER
+    let start x = Timer.delay 1 (fun _ -> Async.StartImmediate x)
+#else
+    let inline start x = Async.Start x
+#endif
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -119,6 +119,7 @@ module Cmd =
             [bind >> start]
 
     module OfAsync =
+        [<Obsolete("Use `AsyncHelpers.start` instead")>]
 #if FABLE_COMPILER
         let start x = Timer.delay 1 (fun _ -> Async.StartImmediate x)
 #else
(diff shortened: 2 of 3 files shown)


src/prelude.fs
```diff
@@ -33,3 +33,10 @@ module internal Timer =
         t.Enabled <- true
         t.Start()
 #endif
+
+module AsyncHelpers =
```
et1975, line 37: Any chance we can call it just 'Async`?

et1975: Also `inline` is the only reason it should remain public, maybe we can remove that and make it `internal`?

MangelMaxime: I originally didn't do it on purpose because I was afraid of shadowing the native API `Async.Start`

But it seems like the native API use `Start` and in our case we use `start` - capital `S` vs lower case `s`

However, I am still a little reluctant to do it, because it could be confusing the user, as if he `open Elmish` which is command then both version will be exposed and it is not obvious which one does what.

MangelMaxime: If we mark it as `internal` then Elmish users will not be able to use it for creating their own `Cmd` helpers.

This is not too much of a concern for .NET, however for Fable it is problematic because they would need to recreate the helper which use `Timer` under the hood.

et1975 (approved):

et1975:
Since this is going to be a breaking change, should we start v5.x branch and merge it there?

MangelMaxime:
Yes we can release it as part of version 5

PR #303, opened 2025-07-11
Implement native support for `Task` and `ValueTask`

xperiandri:


xperiandri:
@et1975 if you have no remarks, I'll clean up the commit history (squash fixups)

Elmish.slnx
et1975, line 1: Not particularly against it, but what does it do for us?

et1975 (approved):
We should probably bump the major since we're bumping not even one but both .NET and Fable.

Elmish.slnx
xperiandri: In the Visual Studio it is easier to work with

websharper/WebSharper.Elmish.fsproj
```diff
@@ -1,9 +1,9 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
xperiandri: Looks like a review glitch. I saw that in AZDO multiple times

src/cmd.fs
```diff
@@ -223,6 +224,132 @@ module Cmd =
...
+                        try
+                            do! task arg
+                        with ex ->
+                            dispatch (ofError ex)
+                    } |> ignore
+                with x ->
+                    dispatch (ofError x)
+            [bind]
+#endif
+#endif
```
MangelMaxime: We could perhaps benefit from a comment to say which condition it is closing.

I am saying that because when reading the diff, I was finding the double `#endif` suspicious

xperiandri: Fixed, good suggestion!

xperiandri:
Also, I reordered commits, so only the breaking changes are at the top

MangelMaxime (approved):

kspeakman (approved):

xperiandri:
Thanks for the approvals! What do we do next? Who merges and releases?

et1975:
I'll need to create a new branch for v5 and retarget the pr, I'll do it
tomorrow when I'm near a computer.

et1975:
Thanks @xperiandri, released as 5.0.0!

src/cmd.fs
```diff
@@ -223,6 +224,132 @@ module Cmd =
...
+                            do! task arg
+                        with ex ->
+                            dispatch (ofError ex)
+                    } |> ignore
+                with x ->
+                    dispatch (ofError x)
+            [bind]
+#endif // WEBSHARPER
+#endif // FABLE_COMPILER
+#if !NETSTANDARD2_0
```
kerams, line 277: This should be `!FABLE_COMPILER || WEBSHARPER`. As it is, v5 fails to compile in Fable when the TFM isn't `netstandard2.0`. That way `OfValueTask` will appear in Intellisense for all versions of the assembly and Fable compilation will fail if the module is referenced (I don't think there's much point in polyfilling the definitions), but `net8.0` projects will then continue to work.

et1975: Would `NET8_0_OR_GREATER && !FABLE_COMPILER || WEBSHARPER` make sense?

et1975: WEBSHARPER can't do ValueTask actually, changing it to `NET8_0_OR_GREATER && !(FABLE_COMPILER || WEBSHARPER)` compiles... 
published as 5.0.1... please make sure it works for your respective use-case.

kerams: It does, thank you.

PR #304, opened 2025-07-13
fix: capture promise exception at runtime and not only at the creation

### LINKED ISSUE #302: Cmd.either broken when error happens
Hi @MangelMaxime,

this change broke our either function:

Image

the try catch doesn't happen on the runtime of promise but only on the "creation of the promise". So when the promise actually runs and an error happens then it is a uncaught javascript exception and everything blows up. Could you please consider to readd the catch statements? 
Want me to send a PR?

Cheers

MangelMaxime:
Fix #302 

I tested with a local reproduction and we do need both the

- `try ... with` => capture exception during promise creation 

	```fs
	let increment model =
	    failwith "dwdw"
	    promise {
	        return model + 1
	    }
	```

- `.catch` => capture exception during promise execution

	```fs
	let increment model =
	    promise {
	        failwith "Error"
	        return model + 1
	    }
	```

### DIFF
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -177,6 +177,7 @@ module Cmd =
                 try
                     (task arg)
                         .``then``(ofSuccess >> dispatch)
+                        .catch(unbox >> ofError >> dispatch)
                         |> ignore
                 with x -> x |> unbox |> ofError |> dispatch
             [bind]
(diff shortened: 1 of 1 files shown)


et1975:
Released as 4.4.0

MangelMaxime:
Thank you

PR #306, opened 2025-07-15
Fixed implicit `Task<'T>` cast to `Task`

xperiandri:
@et1975 I've made a mistake with modifying `OfTask.attempt`. Here is the fix

### DIFF
--- a/src/cmd.fs
+++ b/src/cmd.fs
@@ -261,14 +261,14 @@ module Cmd =
             [bind]
 
         /// Command to call a task and map the error
-        let attempt (task: 'a -> Task)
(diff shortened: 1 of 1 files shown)


src/cmd.fs
```diff
@@ -261,14 +261,14 @@ module Cmd =
             [bind]
 
         /// Command to call a task and map the error
-        let attempt (task: 'a -> Task)
+        let attempt (task: 'a -> #Task)
```
et1975, line 264: anyone know how Fable handles this kind of generic constraint these days?

et1975 (approved):

et1975:
Published as 5.0.2

